PR #5, opened 2017-01-05
Use connection factory instead of context everywhere. Be lazy

makasim:


makasim:
close in favore https://github.com/php-enqueue/enqueue-dev/pull/6

PR #34, opened 2017-03-30
fix fsdriver router topic name

bendavies:
As discussed in gitter.

### DIFF
--- a/pkg/fs/Client/FsDriver.php
+++ b/pkg/fs/Client/FsDriver.php
@@ -173,7 +173,7 @@ class FsDriver implements DriverInterface
     private function createRouterTopic()
     {
         return $this->context->createTopic(
-            $this->config->createTransportRouterTopicName($this->config->getRouterTopicName())
+            $this->config->createTransportQueueName($this->config->getRouterTopicName())
         );
     }
 }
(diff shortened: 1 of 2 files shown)


pkg/enqueue/Client/Config.php
```diff
@@ -154,4 +154,14 @@ public static function create(
             $routerProcessorName ?: 'router'
         );
     }
+
+    /**
+     * @param string $name
+     *
+     * @return string
+     */
+    private function getQueueName($name)
```
makasim: Could you please rename it to `getDestinationName`

bendavies: done

makasim:
I have just realized that the router topic must not contain an app it.

The router topic is a shared one. Several applications may send messages to it and hence talk to each other. It is a ground part of message bus support. 

And since message bus is not supported by `fs` transport we should isolate the change in FsDriver. When a sendToRouter method is called we should instead of a topic create a queue and use it.

bendavies:
no problem, i'll take a look.

makasim:
@bendavies yeap that looks good.

bendavies:
@makasim i'll add a test to cover this case.

bendavies:
@makasim done

makasim:
Thanks!

PR #40, opened 2017-04-05
[client] Add ability to define scope of send message.

makasim:


bendavies:
What is this feature for? Can the docs be updated?

makasim:
The message is sent to a global topic and all subscribed applications get the copy of the message. This change allows you to send a message internally in scope of your current application.

makasim:
@bendavies added an issue

PR #51, opened 2017-04-14
Consume daemon

makasim:
Allows easily to spawn consume workers. Reload them by USR1 signal and
stop correctly when the daemon is stopped.

The snippet shows how you can run 10 consume workers (symfony commands). 

```php
<?php
// bin/consume.php

use Symfony\Component\Process\PhpExecutableFinder;
use Symfony\Component\Process\ProcessBuilder;

require_once __DIR__.'/../vendor/autoload.php';

$phpBin = (new PhpExecutableFinder)->find();
if (false === $phpBin) {
    throw new \LogicException('Php executable could not be found');
}

$daemon = new \App\Infra\Symfony\Daemon();

$builder = new ProcessBuilder([$phpBin, 'bin/console', 'enqueue:consume', '--setup-broker', '-vvv']);
$builder->setPrefix('exec');
$builder->setWorkingDirectory(realpath(__DIR__.'/..'));
$builder->setEnv('MASTER_PROCESS_PID', getmypid());
$daemon->addWorker('consumer', 3, $builder);

$daemon->start();
```

To run execute

```
$ php ./bin/consume.php &; DAEMON_PID=$!
```

To reload workers send a signal

```
$ kill -10 $DAEMON_PID
```

To stop correctly daemon

```
$ kill -15 $DAEMON_PID
```

could be used with a docker

```yaml
# docker-compose.yml

app_consumers:
    image: 'formapro/nginx-php-fpm:latest-all-exts'
    restart: 'always'
    entrypoint: "php bin/consumers.php"
    working_dir: '/app'
    depends_on:
      - 'rabbitmq'
      - 'mongo'
    volumes:
      - './:/apps/jm:cached'
```


makasim:
The real life test show that the approach is much much slower than a pervious approach we used (run a consumer coomand as docker entrypoint). Should be fixed before merge

makasim:
I've updated the code. it must work better now but I haven't checked.

bendavies:
btw, have you seen https://github.com/Seldaek/signal-handler/?

makasim:
@bendavies no, I've not seen it. Enqueue manages signals in the dedicated extension which is not enabled on windows. 

This daemon thing will be released as a standalone package and correct work of pcntl is critical for the daemon.

Issue #32, opened 2017-03-29
Additional drivers


bendavies:
Hey!

This library looks awesome, nice work!
I'd really like to use it to replace our existing implementation, but we currently use SQS and a database as Transports. Are there any plans to support SQS/Database as Transports in enqueue, or would PR's be considered?

Thanks!

makasim:
There is a plan but to be honest I would not expect it any time soon. We use RabbitMQ via AMQP so there is no need for SQS at the moment. 

The database transport could be taken from https://github.com/orocrm/OroMessageQueueComponent/tree/master/Transport/Dbal. It has to be adjust but at least no need to write it from scratch.

Of course you can work on a PR (I'd very happy to merge it) or you can create a standalone extension (I am fine with it too).

makasim:
@bendavies In dbal transport we tried to support delivery guaranty (If a consumer fails the message is get back to the queue and marked redelivered). 

That was hardest part. The currently visible solution is no go solution. It simply returns messages to queue after some period of time 2 minutes. So if a consumer takes more that that time to process a message you are in trouble. 

We worked on another more sophisticated solution but they have not even merged that to master yet. 

I would suggest to completely drop  the support of the feature for DBAL transport. It makes it simpler, less deadlocks, better performance and so on.

PR #60, opened 2017-04-25
Amazon SQS Transport

ASKozienko:


pkg/sqs/SqsContext.php
```diff
@@ -0,0 +1,196 @@
...
+            'QueueUrl' => $this->getQueueUrl($dest->getQueueName()),
+        ]);
+
+        unset($this->queueUrls[$dest->getQueueName()]);
+    }
+
+    /**
+     * @param SqsDestination $dest
+     */
+    public function purgeQueue(SqsDestination $dest)
```
makasim: call the method `purge`

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/32

Issue #64, opened 2017-04-27
[amqp] Get message count


J7mbo:
In `AMQPContext::declareQueue()`, calling [`$extQueue->declareQueue()`](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-ext/AmqpContext.php#L130) returns an integer which is the message count. But this isn't returned anywhere by the initial method.

Is my only option to override this context with my own, and copy / paste the `declareQueue()` method code, adding a return for the call through to `AMQP::declareQueue()`? Can I provide a PR? Or is there another way to get a message count from a queue? Thanks!

makasim:
>  Can I provide a PR?

Yes sure. 

Just out of curiosity what is your usecase? Why do you need a number of messages and how do you use that?

J7mbo:
I'm limiting the amount of jobs a given user in my application can add to a queue at any one time. Thanks! Will add a PR soon.

grekpg:
Im not understatnd, how i can get messages queue -  i use symfony bundle. I try 
```

    $factory = new AmqpConnectionFactory('amqp+ext://rabbitmq:rabbitmq@rabit_net:5672/%2f');

        $context = $factory->createContext();

        $fooTopic = $context->createTopic(EnqueueProducer::TOPIC_CREATE_PRODUCT);

        $count = $context->declareTopic($fooTopic);

        $queue = $context->createQueue('aQueue');

        $consumer = $context->createConsumer($fooTopic);
```

where i can count messages ?

makasim:
```
$queue = $context->createQueue('aQueue');
$count = $context->declareQueue($queue);
```

topic (or in terms of rabbitmq -exchange) does not store messages

grekpg:
```
$factory = new AmqpConnectionFactory('amqp+ext://rabbitmq:rabbitmq@rabit_net:5672/%2f');

        $context = $factory->createContext();

       $queue = $context->createQueue('enqueue.app.default');
```

error:

`Server channel error: 406, message: PRECONDITION_FAILED - inequivalent arg 'durable' for queue 'enqueue.app.default' in vhost '/': received 'false' but current is 'true'
`
```
  $queue = $context->createQueue('enqueue.app.default');

        $queue->setArgument('durable',true);

        $count = $context->declareQueue($queue);
```

same error. 
Can i get queue from some Symfony service ? or how set durable true (this is default settting from symfony) i just need count messages .. 


[...]

PR #66, opened 2017-04-28
Calling AmqpContext::declareQueue() now returns an integer holding the queue message count

J7mbo:
Calling AmqpContext::declareQueue() now returns an integer holding the queue message count.

See #64.


makasim:
Could you please add a unit test?

PR #75, opened 2017-05-09
Multi Transport Simple Client

ASKozienko:


makasim:
That looks good to me. Could you please create a new package for the simple client ( we can call it `simple-client`). So the dependency on symfony container component is explicitly defined in composer.json.

makasim:
Add other factories and move tests that depends on simple client where appropriate

pkg/enqueue/composer.json
```diff
@@ -21,7 +21,8 @@
         "symfony/dependency-injection": "^2.8|^3",
         "symfony/config": "^2.8|^3",
         "enqueue/null": "^0.3",
-        "enqueue/test": "^0.3"
+        "enqueue/test": "^0.3",
+        "enqueue/simple-client": "^0.3"
```
makasim: IMO enqueue should not require simple client. Why do you think we should do?

pkg/simple-client/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2016 Kotliar Maksym
```
makasim: 2017

pkg/simple-client/SimpleClient.php
```diff
@@ -1,19 +1,31 @@
...
+use Enqueue\Redis\Symfony\RedisTransportFactory;
+use Enqueue\Sqs\Symfony\SqsTransportFactory;
+use Enqueue\Stomp\Symfony\RabbitMqStompTransportFactory;
+use Enqueue\Stomp\Symfony\StompTransportFactory;
 use Enqueue\Symfony\DefaultTransportFactory;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
 /**
- * Experimental class. Use it speedup setup process and learning but consider to switch to custom solution (build your own client).
+ * Use it speedup setup process and learning but consider to switch to custom solution (build your own client).
```
makasim: remote it completly

pkg/simple-client/composer.json
```diff
@@ -0,0 +1,43 @@
...
+        {
+            "type": "vcs",
+            "url": "git@github.com:php-enqueue/test.git"
+        }
+    ],
+    "require": {
+        "php": ">=5.6",
+        "enqueue/enqueue": "^0.3",
+        "symfony/dependency-injection": "^2.8|^3",
+        "symfony/config": "^2.8|^3"
```
makasim: add `symfony/console`.

pkg/simple-client/composer.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+    "require-dev": {
+        "phpunit/phpunit": "~5.5",
+        "enqueue/test": "^0.3"
+    },
+    "suggest": {
```
makasim: remote transport suggestions. (they are suggeted by enqueue/enqueue) lib

makasim:
Thank you @ASKozienko

PR #78, opened 2017-05-11
[fs] Filesystem transport must create a storage dir if it does not exists.

makasim:
fixes https://github.com/symfony/recipes-contrib/pull/16#discussion_r116027232


robfrawley:
I don't think you meant to close symfony/recipes-contrib#16 when you closed this.

PR #97, opened 2017-05-23
Reuse json util where it possible

Koc:


makasim:
Transports is a low level abstraction and enqueue works **on top of it**. `enqueue\enqueue` contains a lot of features working on top of it, like RPC, the easy to use client, consumption simplifications, cli commands. 

Besides it does not make much sense to me to require the whole that stuff just to be able to reuse JSON utill.

Koc:
sounds reasonable, agree with you

PR #102, opened 2017-05-24
[POC] add message trait, add get\set dest to message interface

makasim:


makasim:
@ASKozienko I need your explicit approve on this. Please do review.

makasim:
After internal discussion, the trait will be dropped, the spec test remains.

Issue #87, opened 2017-05-16
Transports must support configuration via DSN string


makasim:
The list of transport that could support DSN:

* [x] amqp-ext
* [x] amqp-bunny
* [x] amqp-lib
* [x] null
* [x] fs
* [ ] redis
* [ ] stomp
* [x] doctrine dbal
* [ ] amazon sqs
* [x] beanstalk
* [x] gearman
* [x] google pub sub (gps)

PR #104, opened 2017-05-25
[dbal] Add DSN support.

makasim:
ref https://github.com/php-enqueue/enqueue-dev/issues/87


Koc:
How it possible to reuse existing connection?

makasim:
@Koc use this option https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/dbal/Symfony/DbalTransportFactory.php#L50.

PR #106, opened 2017-05-30
Factories rework.

makasim:
Here, I tried to solve next problems:

* Exception: Incompatible use of dynamic environment variables "ENQUEUE_DSN" found in parameters.
* Configure several transports
* Configure transports in runtime (to fix the exception).


makasim:
I am not sure this is the right way to go so I put it on hold and will do a quick fix for the exception only.

PR #114, opened 2017-06-13
RPC Deletes Reply Queue After Receive Message

ASKozienko:


makasim:
Could you please add `getMessage` method back and mark it as deprecated?

Issue #128, opened 2017-07-04
Amqp lib support


pavellagoda:
I am going to use enqueue in our next project but I cannot use amqp extension. Are you planning to support phpamqplib?

makasim:
I wish we could do it but we cannot because of a contact.
But you can do it on your own. That's not that hard and if you have any questions don't hesitate to ask them here or by email or skype.

PR #136, opened 2017-07-18
Add AmqpLib support

fibula:
fixes https://github.com/php-enqueue/enqueue-dev/issues/128

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,6 +8,7 @@
     "enqueue/stomp": "*@dev",
     "enqueue/amqp-ext": "*@dev",
     "enqueue/amqp-lib": "*@dev",
+    "php-amqplib/php-amqplib": "^2.7@dev",
     "enqueue/redis": "*@dev",
     "enqueue/fs": "*@dev",
     "enqueue/null": "*@dev",
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "enqueue/enqueue": "*@dev",
     "enqueue/stomp": "*@dev",
     "enqueue/amqp-ext": "*@dev",
-    "enqueue/amqplib": "*@dev",
+    "enqueue/amqp-lib": "*@dev",
     "enqueue/redis": "*@dev",
     "enqueue/fs": "*@dev",
     "enqueue/null": "*@dev",
@@ -71,7 +71,7 @@
     },
     {
       "type": "path",
-      "url": "pkg/amqplib"
+      "url": "pkg/amqp-lib"
     },
     {
       "type": "path",
--- a/composer.json
+++ b/composer.json
@@ -7,6 +7,7 @@
     "enqueue/enqueue": "*@dev",
     "enqueue/stomp": "*@dev",
     "enqueue/amqp-ext": "*@dev",
+    "enqueue/amqplib": "*@dev",
     "enqueue/redis": "*@dev",
     "enqueue/fs": "*@dev",
     "enqueue/null": "*@dev",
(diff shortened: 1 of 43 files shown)


composer.json
```diff
@@ -7,6 +7,7 @@
     "enqueue/enqueue": "*@dev",
     "enqueue/stomp": "*@dev",
     "enqueue/amqp-ext": "*@dev",
+    "enqueue/amqplib": "*@dev",
```
makasim: Could you please rename it to `amqp-lib`? Same goes for the composer package name

pkg/amqplib/AmqpConnectionFactory.php
```diff
@@ -0,0 +1,49 @@
...
+use Interop\Queue\PsrConnectionFactory;
+use PhpAmqpLib\Connection\AbstractConnection;
+use PhpAmqpLib\Connection\AMQPStreamConnection;
+
+class AmqpConnectionFactory implements PsrConnectionFactory
+{
+    private $config;
+    private $connection;
+
+    public function __construct(array $config = [])
```
makasim: can you please add doc blocks for the properties and methods

pkg/amqplib/AmqpConsumer.php
```diff
@@ -0,0 +1,79 @@
...
+use Interop\Queue\PsrConsumer;
+use Interop\Queue\PsrMessage;
+use PhpAmqpLib\Channel\AMQPChannel;
+use PhpAmqpLib\Message\AMQPMessage as LibAMQPMessage;
+use PhpAmqpLib\Wire\AMQPTable;
+
+class AmqpConsumer implements PsrConsumer
+{
+    private $channel;
+    private $queue;
```
makasim: we use php-cs-fixer to make sure the code looks similar. Could you please use it too? https://github.com/php-enqueue/enqueue-dev/blob/master/docs/contribution.md

pkg/amqplib/AmqpConsumer.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public function getQueue()
+    {
+        return $this->queue;
+    }
+
+    public function receive($timeout = 0)
+    {
+        $end = microtime(true) + ($timeout / 1000);
```
makasim: it would be good to add a `basic_consume` support. But it could be done at a later stage.

pkg/amqplib/AmqpConnectionFactory.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+namespace Enqueue\Amqplib;
```
makasim: Could you please change it to `Enqueue\AmqpLib`?

makasim:
@fibula It would be also great if you could add some tests. You can use `queue-spec` at least. It should not be hard to add them

https://github.com/queue-interop/queue-spec

pkg/amqp-lib/AmqpConsumer.php
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * @param int $timeout
+     *
+     * @return AmqpMessage|null
+     */
+    public function receive($timeout = 0)
+    {
+        $end = microtime(true) + ($timeout / 1000);
+
+        while (0 === $timeout || microtime(true) < $end) {
```
makasim: would be able to add a support of `basic_consume`. You can look how we did it in amqp ext transport

ASKozienko: ok I will make a try

fibula: ok, I will make a try

PR #138, opened 2017-07-24
Amqp Interop

ASKozienko:


pkg/amqp-ext/Client/AmqpDriver.php
```diff
@@ -3,14 +3,14 @@
 namespace  Enqueue\AmqpExt\Client;
```
makasim: @ASKozienko could you please move amqp client related to Enqueue\Client\Amqp namespace?

makasim: Please do it as a standalone PR.

PR #167, opened 2017-08-15
Google Pub/Sub

ASKozienko:


pkg/gps/Symfony/GpsTransportFactory.php
```diff
@@ -0,0 +1,130 @@
...
+        $builder
+            ->beforeNormalization()
+            ->ifString()
+                ->then(function ($v) {
+                    return ['dsn' => $v];
+                })
+            ->end()
+            ->children()
+                ->scalarNode('dsn')
+                    ->info('The connection to AMQP broker set as a string. Other parameters are ignored if set')
```
makasim: amqp is not correct in this context

pkg/gps/composer.json
```diff
@@ -0,0 +1,41 @@
+{
+    "name": "enqueue/gps",
+    "type": "library",
+    "description": "Message Google Cloud Pub/Sub Transport",
+    "keywords": ["messaging", "queue", "google", "pubsub"],
+    "license": "MIT",
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "git@github.com:php-enqueue/test.git"
```
makasim: there is no need in defining it explicitly 

add enqueue/test to require-dev

PR #175, opened 2017-08-22
Add FIFO support to Amazon SQS package

studioromeo:
Allows for FIFO queues on Amazons SQS platform to be used. These queues have two characteristics that needed to be addressed in the adapter:

1. FIFO queues have a suffix of `.fifo`
2. Each message should have a `messageGroupId` and a `messageDeduplicationId`

In order to define a queue as FIFO I apply the suffix to the queue name eg
```yml
enqueue:
    client:
        default_processor_queue: default.fifo
```

The messageGroupId is a md5 hash of the method name that created the `$transportMessage` object and the `messageDeduplicationId` is a md5 hash of the body of the message.

Question:
I notice in SqsDestination that there are methods `setFifoQueue` & `setContentBasedDeduplication`. Should these be set as part of `createQueue`?

### DIFF
--- a/pkg/sqs/Client/SqsDriver.php
+++ b/pkg/sqs/Client/SqsDriver.php
@@ -85,7 +85,7 @@ class SqsDriver implements DriverInterface
     public function createQueue($queueName)
     {
         $transportName = $this->queueMetaRegistry->getQueueMeta($queueName)->getTransportName();
-        $transportName = str_replace('.', '_dot_', $transportName);
+        $transportName = preg_replace('/\.(?!fifo)/', '_dot_', $transportName);
 
         return $this->context->createQueue($transportName);
     }
--- a/pkg/sqs/Client/SqsDriver.php
+++ b/pkg/sqs/Client/SqsDriver.php
@@ -135,6 +135,11 @@ class SqsDriver implements DriverInterface
         $transportMessage->setReplyTo($message->getReplyTo());
         $transportMessage->setCorrelationId($message->getCorrelationId());
 
+        if (preg_match('/(\.fifo)$/', $message->getProperty(Config::PARAMETER_PROCESSOR_QUEUE_NAME))) {
+            $transportMessage->setMessageGroupId(md5(__METHOD__));
+            $transportMessage->setMessageDeduplicationId(md5($message->getBody()));
+        }
+
         return $transportMessage;
     }
 
--- a/pkg/sqs/Tests/Client/SqsDriverTest.php
+++ b/pkg/sqs/Tests/Client/SqsDriverTest.php
@@ -62,6 +62,28 @@ class SqsDriverTest extends TestCase
         $this->assertSame('aQueueName', $queue->getQueueName());
     }
 
+    /**
+     * @group fifo
+     */
+    public function testShouldCreateAndReturnAFIFOQueueInstance()
+    {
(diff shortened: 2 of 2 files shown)


makasim:
LGTM

studioromeo:
@makasim Should I set `setFifoQueue` & `setContentBasedDeduplication` as part of `createQueue` ? I'm not clear on how these get used

bendavies:
I'm planning on using this eventually so I'm having a poke around too.

It seem https://github.com/php-enqueue/enqueue-dev/blob/2c785fef478e7a4c3b9258af1fcde8aa179cb5e7/pkg/sqs/SqsDestination.php#L156-L162
should be called somewhere?

According to
https://github.com/php-enqueue/enqueue-dev/blob/2c785fef478e7a4c3b9258af1fcde8aa179cb5e7/pkg/sqs/SqsDestination.php#L165-L193
It seems `MessageDeduplicationId`isn't needed   as long as `ContentBasedDeduplication` is set, so that method should be called somewhere too?`

makasim:
@ASKozienko ping

ASKozienko:
Hi, as I see here is slight misunderstanding. This solution will not work, and I don't think SQS FIFO is good for enqueue client, and I don't think we can guarantee it will work as expected with enqueue client. I mean we can't guarantee FIFO.
A bit of theory:
Enqueue has 2 layers:
1. transport - this is a tiny abstraction over broker API or broker php lib or broker php ext etc. Transport SHOULD support all features broker has. 
2. client  - this is a high level lib. It makes possible work with queues as simple as with events. As user you send message to client topic (not to transport topic) and receive message in one or many processors. Is not your deal to know how to create queues, exchanges, topics, messages etc on broker side. Client does whole this work through the middleware called `DriverInterface`. 
Now I see next mistakes:
1. For FIFO SQS requires queue name ends with ".fifo" and attribute "FifoQueue="true"".  Use next method to set this attribute SqsDestination::setFifoQueue(true).
2. Enqueue client has router. Router has message destination map (topic->processor(s) map) and router makes message duplicates for every destination. It means queue will have 2 or more messages with same `MessageDeduplicationId ` and SQS will delete second message, you just lose messages. To fix it you have to use separate queues for router and for every processor. Or you have to generate unique id for every message in `SqsDriver::sendToRouter` and `SqsDriver::sendToProcessor`. From SQS doc:
Message deduplication applies to an entire queue, not to individual message groups . Maybe `SqsDestination::setContentBasedDeduplication(true)` will fix this problem too. Because enqueue client changes message headers and SQS message this is json serialized string which contains not only message body but properties and headers too.

makasim:
My two cents:

1. It might cause problems if there are many consumers receive messages from the FIFO queue. So if you use FIFO you have to have exactly one consumer. 
2. It is hard to predict how it would work with the router and other processors working in the same queue. 
3. It is hard to predict how it would work with a delay extension. It puts the message back to queue.

ASKozienko:
expression `/\.(?!fifo$)/` is better as `.fifo` must be at the end.

studioromeo:
I think based on your responses to this PR that this is not an acceptable direction to take with this adapter so I'll close this PR. Thanks for the review 👍

makasim:
@studioromeo I suggest you using sqs transport directly.

Issue #170, opened 2017-08-16
[FS] Frame not being read correctly


jenkoian:
Not sure if this is a bug or if I've done something not quite right, but I get the following error when certain messages are attempted to be consumed:

```
[LogicException]
  Cannot decode json message 'JSON HERE'
```

Digging into it, the frames seem to be getting split *between* a `|` and a `{` meaning the following isn't getting hit and thus the message isn't read from the file correctly. 

https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/fs/FsConsumer.php#L172

Is there a reason 64 was chosen? Was it just a random number and I've been desperately unlucky?

makasim:
Just a random number, The message extraction logic is much much simpler if the message length is predictable

makasim:
I've never run in this but it could still be a bug.

jenkoian:
Can confirm that if I change my message by a few characters it works as expected, so guess I was just unlucky, but yeah sounds like a bug.

makasim:
Could you post a message that causes it?

jenkoian:
This is the one I was running into issues with, changed ids to x's but kept the length the same. If you change `unsubscribe` to `unscribe` it then works ok.

```json
{
	"body": "{\"consumerId\":\"user:xxxxxx\",\"producerId\":\"user:xxxxxx\",\"type\":\"unsubscribe\"}",
	"properties": {
		"enqueue.topic_name": "__command__",
		"enqueue.command_name": "updateConsumerNetwork",
		"enqueue.processor_name": "enqueue.client.router_processor",
		"enqueue.processor_queue_name": "default"
	},
	"headers": {
		"content_type": "text\/plain",
		"message_id": "9cb6dcb9-8b6c-402a-ae61-3cd6b321c150",
		"timestamp": 1502689812,
		"reply_to": null,
		"correlation_id": ""
	}
}
```

jenkoian:
What about a few file for each message? I *think* that's what bernard does but haven't had time to look into properly https://github.com/bernardphp/bernard/blob/master/src/Driver/FlatFileDriver.php

makasim:
It failed on message size equal to 24

```
Received: a
Sent: 2
Received: aa
Sent: 3
Received: aaa
Sent: 4
Received: aaaa
Sent: 5
Received: aaaaa
Sent: 6
Received: aaaaaa
Sent: 7
Received: aaaaaaa
Sent: 8
Received: aaaaaaaa
[...]

PR #176, opened 2017-08-24
Use Query Builder for better support across platforms.

jenkoian:
Not all drivers supported FOR UPDATE queries so this change changes the
raw SQL to use the query builder thus allowing DBAL to construct the
appropriate query per platform. It then grabs the SQL and appends the
read lock SQL but grabs the SQL part from the correct platform as
to not break portability.

### DIFF
--- a/pkg/dbal/DbalConsumer.php
+++ b/pkg/dbal/DbalConsumer.php
@@ -169,12 +169,18 @@ class DbalConsumer implements PsrConsumer
         try {
             $now = time();
 
-            $sql = sprintf(
-                'SELECT * FROM %s WHERE queue=:queue AND '.
-                '(delayed_until IS NULL OR delayed_until<=:delayedUntil) '.
-                'ORDER BY priority DESC, id ASC LIMIT 1 FOR UPDATE',
-                $this->context->getTableName()
-            );
+            $query = $this->dbal->createQueryBuilder();
+            $query
+                ->select('*')
+                ->from($this->context->getTableName())
+                ->where('queue = :queue')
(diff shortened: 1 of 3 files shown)


makasim:
@jenkoian what platform are you going to use?

jenkoian:
@makasim I'm running into https://github.com/php-enqueue/enqueue-dev/issues/170 all the time on dev, so was hoping to use SQLite.

makasim:
FYI: It has not ever tested agains SQLite. We used it with mysql and postgresql. 

I'll try to find some time next week to fix #170.

makasim:
It looks good but could you please look at travis, it complains about something. 
Ignore failed test if they are kafka related

makasim:
`FOR UPDATE` is a write lock, and you change it to [getReadLockSQL](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/MySqlPlatform.php#L1010) which is not the same.

jenkoian:
@makasim ok will change to getWriteLockSQL

jenkoian:
@makasim updated to use getReadLockSQL and ran the cs fixer to hopefully fix the build.

jenkoian:
@makasim the remaining build failure seem to be SQS related, possibly to do with this warning? ```WARNING: The ENQUEUE_AWS__SQS__KEY variable is not set. Defaulting to a blank string.```

makasim:
@jenkoian that ok. travis does not add secret envs to a build if a build is initialized by a not authorized person.

Issue #187, opened 2017-08-29
Serialization beyond JSON


akucherenko:
Current implementation of RdKafka transport supports a single way to serialize/unserialize payloads, that is either `json_encode` of `PsrMessage` [RdKafkaProducer.php](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/RdKafkaProducer.php#L40) or `json_decode` and further `PsrMessage` population [RdKafkaConsumer.php](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/RdKafkaConsumer.php#L154). In this case, consumer and producer are only mutually compatible and consuming messages of different format is hardly possible. E.g. no chance to plug into an Avro-encoded stream.

An option here might be making serializer configurable. Or provide abstraction around transforming payload into an object. Maybe there are unrecognized reasons behind current implementation. Anyway, I'd like to discuss that and provide help if needed.

PR #188, opened 2017-08-30
[rdkafka] Add abilito change the way a message is serialized.

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/187

TODO:

* ~~upd eqnueue bundle~~
* [x]  add docs


makasim:
@ASKozienko @akucherenko could you please take a look.

PR #190, opened 2017-09-05
[Symfony] added toolbar info in profiler

Miliooo:
I've updated the profiler panel
- Shows an icon + message count **when** messages have been sent
- Added an icon in the profiler menu

**Preview**
rsz_symf_toolbar

I'm not claiming it's the best icon ever but now we don't have to open the profiler page to check how many messages were sent


makasim:
Thanks for the contribution.

Issue #191, opened 2017-09-05
Redis consumer has very high resource usage


JunkMyFunk:
I have configured the Symfony bundle to use Redis via php-redis. The queue producer and consumer are working, however when I run the console command `php app/console enqueue:consume --setup-broker` it seems to poll Redis every few milliseconds causing the PHP script to have about 50% CPU usage and Redis about 20%.

Is there an option to set a timeout to reduce the polling rate and reduce resource usage?

```
PHP 5.6
Redis server 2.8.4
symfony/symfony: 2.8.27
enqueue/async-event-dispatcher: 0.7.5 
enqueue/enqueue: 0.7.9
enqueue/enqueue-bundle: 0.7.8
enqueue/redis: 0.7.7
```

makasim:
There is no such option yet. Here's a solution we made for dbal transport. It might work for redis too.


https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/dbal/DbalConsumer.php#L96

makasim:
The other solution: Add the third argument to queue consumer which is idle time. https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue-bundle/Resources/config/client.yml#L82

makasim:
There is `--receive-timeout` option is available since 0.7.11. You can use it to increase time a consumer waits on a socket for a message.

PR #193, opened 2017-09-08
Queue Consumer Options

ASKozienko:
fixes https://github.com/php-enqueue/enqueue-dev/issues/191


pkg/enqueue-bundle/Tests/Unit/DependencyInjection/ConfigurationTest.php
```diff
@@ -423,4 +423,25 @@ public function testShouldAllowEnableAsyncEvents()
             ],
         ], $config);
     }
+
+    public function testShouldAllowConfigureConsumption()
```
makasim: there must be test for such case: "nothing is set then default values are used"

pkg/enqueue-bundle/Tests/Unit/DependencyInjection/EnqueueExtensionTest.php
```diff
@@ -535,4 +535,28 @@ public function testShouldNotAddJobQueueEntityMappingIfDoctrineBundleIsNotRegist
...
+    {
+        $container = new ContainerBuilder();
+
+        $extension = new EnqueueExtension();
+        $extension->load([[
+            'client' => [],
+            'transport' => [
+            ],
+            'consumption' => [
+                'idle-timeout' => 123,
```
makasim: it must be underscore "_" instead of dash "-"

pkg/enqueue/Symfony/Consumption/QueueConsumerOptionsCommandTrait.php
```diff
@@ -0,0 +1,36 @@
...
+trait QueueConsumerOptionsCommandTrait
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configureQueueConsumerOptions()
+    {
+        $this
+            ->addOption('idle-timeout', null, InputOption::VALUE_REQUIRED, 'The time in milliseconds queue consumer waits if no message received.')
+            ->addOption('receive-timeout', null, InputOption::VALUE_REQUIRED, 'The time in milliseconds queue consumer waits for a message.')
```
makasim: those values should not be required imo

PR #194, opened 2017-09-11
fixed NS

chdeliens:
was causing "PHP Fatal error:  Cannot use Interop\Amqp\AmqpConsumer as AmqpConsumer because the name is already in use in enqueue/amqp-ext/Flags.php on line 5"

### DIFF
--- a/pkg/amqp-ext/Flags.php
+++ b/pkg/amqp-ext/Flags.php
@@ -2,11 +2,11 @@
 
 namespace Enqueue\AmqpExt;
 
-use Interop\Amqp\AmqpConsumer;
-use Interop\Amqp\AmqpDestination;
-use Interop\Amqp\AmqpMessage;
-use Interop\Amqp\AmqpQueue;
-use Interop\Amqp\AmqpTopic;
+use Interop\Amqp\AmqpConsumer as InteropAmqpConsumer;
+use Interop\Amqp\AmqpDestination as InteropAmqpDestination;
+use Interop\Amqp\AmqpMessage as InteropAmqpMessage;
+use Interop\Amqp\AmqpQueue as InteropAmqpQueue;
+use Interop\Amqp\AmqpTopic as InteropAmqpTopic;
 
 class Flags
 {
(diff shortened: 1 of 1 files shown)


pkg/amqp-ext/Flags.php
```diff
@@ -2,11 +2,11 @@
...
-use Interop\Amqp\AmqpConsumer;
-use Interop\Amqp\AmqpDestination;
-use Interop\Amqp\AmqpMessage;
-use Interop\Amqp\AmqpQueue;
-use Interop\Amqp\AmqpTopic;
+use Interop\Amqp\AmqpConsumer as InteropAmqpConsumer;
+use Interop\Amqp\AmqpDestination as InteropAmqpDestination;
+use Interop\Amqp\AmqpMessage as InteropAmqpMessage;
+use Interop\Amqp\AmqpQueue as InteropAmqpQueue;
+use Interop\Amqp\AmqpTopic as InteropAmqpTopic;
```
makasim: Shouldn't we get rid of these imports and use `Enqueue\AmqpExt\AmqpConsumer` instead.

chdeliens: You could, but the same I did above has been done in other classes.
I suggest to keep it that way for now and refactor mid-term. What do you think?

chdeliens:
I can reproduce on PHP 5.6.31 by running examples/consume.php.

pkg/amqp-ext/Flags.php
```diff
@@ -2,11 +2,11 @@
 
 namespace Enqueue\AmqpExt;
 
-use Interop\Amqp\AmqpConsumer;
```
makasim: That would be enought to remove this single line `use Interop\Amqp\AmqpConsumer;`. It must work.

chdeliens: Since AmqpConsumer from Enqueue implements InteropAmqpConsumer, it should work indeed! I'm gonna give it a quick test and keep you posted.

chdeliens: Works! Confirmed by testing on 5.6.31. Should work on other versions as well. let's go for it :)

makasim: @chdeliens great! Could you please update the PR?

PR #195, opened 2017-09-11
Fixed losing message priority for dbal driver

vtsykun:


makasim:
LGTM in general though there is issue with code style https://travis-ci.org/php-enqueue/enqueue-dev/jobs/274379202

makasim:
@vtsykun could you please add this spec test https://github.com/queue-interop/queue-spec/blob/master/src/SendAndReceivePriorityMessagesFromQueueSpec.php

You can find out how to do it in other transport which use spec tests. For example [amqp ext](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-ext/Tests/Spec/AmqpSendAndReceivePriorityMessagesFromQueueTest.php).

pkg/dbal/DbalConsumer.php
```diff
@@ -176,7 +176,7 @@ protected function receiveMessage()
                 ->where('queue = :queue')
                 ->andWhere('(delayed_until IS NULL OR delayed_until <= :delayedUntil)')
                 ->orderBy('priority', 'desc')
-                ->orderBy('id', 'asc')
+                ->addOrderBy('id', 'asc')
```
makasim: it should be `desc`. gonna fix it

makasim: for ref https://github.com/php-enqueue/enqueue-dev/pull/198/files#diff-12c1fdd4a765fda7186569d999b34749R179

PR #198, opened 2017-09-13
[dbal] add priority support on transport level.

makasim:
It also fixes ordering issue.


pkg/dbal/DbalConsumer.php
```diff
@@ -176,7 +176,7 @@ protected function receiveMessage()
                 ->where('queue = :queue')
                 ->andWhere('(delayed_until IS NULL OR delayed_until <= :delayedUntil)')
                 ->orderBy('priority', 'desc')
-                ->addOrderBy('id', 'asc')
+                ->addOrderBy('id', 'desc')
```
jenkoian: Won't this change it from a fifo queue to a lifo queue?

makasim: I messed up everything. I must've fixed priority ordering not id.

PR #207, opened 2017-09-27
Fixes the notation for Twig template names in the data collector

Lctrs:
Fix the error below when attempting to open the profiler in a Symfony Flex project : 

`The profiler template "EnqueueBundle:Profiler:panel.html.twig" for data collector "enqueue.message_queue" does not exist.`

### DIFF
--- a/pkg/enqueue-bundle/Resources/config/client.yml
+++ b/pkg/enqueue-bundle/Resources/config/client.yml
@@ -136,7 +136,7 @@ services:
         tags:
             -
               name: 'data_collector'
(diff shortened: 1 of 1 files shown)


makasim:
Is it BC break. Do old Symfony versions still work?

Lctrs:
This notation is present (and recommended) in the 2.8 version of the Symfony docs. So yes.

Lctrs:
https://symfony.com/doc/2.8/templating/namespaced_paths.html

makasim:
@Lctrs thanks for the clarification.

PR #222, opened 2017-10-12
Adds to small extension points to JobProcessor

iainmckay:


makasim:
@iainmckay Could you shade some light on how this extension point is supposed to be used?

iainmckay:
Sorry, yes. 

My team is currently in the middle of modernising an application. One aspect we are in the process of updating is how we process some of our workloads and pushing them to a background/asynchronous workflow. We're using the `job-queue` to keep track of what is running, enforce uniqueness, etc.

New functionality built for this application was designed to work in a completely async fashion and to talk with our command bus but we have many older parts of the application which, for various reasons, just don't play well and assumes all operations are executing in the same process. We override the `JobProcessor` to decorate the `Job` in the new `saveJob` method and we suppress the sending of the event in `sendCalculateRootJobStatusEvent` so that it does not communicate with the message bus at all. This allows us to use the `job-queue` for everything regardless of it is in the legacy part of our application or in the shiny new part. 

This was the least hassle way we came up with to bridge our legacy code. I would say it is not a massive change in itself but it helps us out massively.

I would expect this is very particular to our project and even for ourselves it has a limited lifespan as we're rewriting the ageing modules.

makasim:
@iainmckay Fair enough. Could you please add a link to your description in the docblocks? I am going to merge it after.

iainmckay:
If I've understood you correctly, I've updated the PR.

Thanks for merging it.

makasim:
https://github.com/php-enqueue/job-queue/releases/tag/0.7.19

makasim:
@iainmckay As a side note. We are working on a job manager microservice. It is almost done. 

https://github.com/php-comrade/comrade-dev

The doc still needs polishing.If you find it interesting feel free to ask me for guidance

iainmckay:
Thanks. We'll keep an eye on it. 

We're still figuring out exactly how our job system should look as we run a single-tenant app but installed multiple times for our clients which we host so we're trying to keep all data isolated but have a shared worker environment (each worker should stand up the client's instance with their configuration, etc. to execute a job).

makasim:
Comrade does not perform jobs itself, instead it can send a message to queue or http request. So it is perfectly fine to have workers on client side (as long as they are connected to a queue or expose http api).

Worth to mention that a worker could be written in any language not only PHP.

PR #241, opened 2017-10-21
Only add Ampq transport factories when packages are found

jverdeyen:


makasim:
AmqpTransportFactory is in `enqueue/enqueue` package which is always required by the bundle. 
The factory inside does class_exists checks and uses only what is available.

jverdeyen:
But the `RabbitMqAmqpTransportFactory` uses `DelayStrategyTransportFactoryTrait` which I am no using with the following reqs:

```
"enqueue/enqueue-bundle": "<=1.0.0",
"enqueue/job-queue": "<=1.0.0",
"enqueue/sqs": "<=1.0.0",
"enqueue/redis": "<=1.0.0",
"enqueue/dbal": "<=1.0.0",
```

Should I always require `enqueue/enqueue` also ?

makasim:
You shouldn't, the bundle heavily [rely on it](https://github.com/php-enqueue/enqueue-bundle/blob/master/composer.json#L10)

makasim:
Are there any errors you face?

jverdeyen:
The [DelayStrategyTransportFactoryTrait](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-tools/DelayStrategyTransportFactoryTrait.php) is from the [ampq-tools package](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-tools/composer.json). Which isn't a requirement for [enqueue/enqueue](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/composer.json) nor for [the enqueue-bundle](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue-bundle/composer.json). Or am I missing something? I get an error that the Trait can't be found. Since the bundle tries to load the `RabbitMqAmqpTransportFactory` which uses the `DelayStrategyTransportFactoryTrait`.

makasim:
Okay, let's merge this then. Could you please CS issues please?

jverdeyen:
This is the correct check to avoid adding the TransportFactories?

https://github.com/php-enqueue/enqueue-dev/pull/241/files#diff-21599724c6f8c888b53103436890a85fR60

makasim:
> This is the correct check to avoid adding the TransportFactories?

yeap

makasim:
Commit contains better fix for this issue https://github.com/php-enqueue/enqueue-dev/commit/031380e5f8b9069dcf9d996da42c58ead38e7ac3. I kept both changes for now, though  ones you added are not needed any more

PR #242, opened 2017-10-26
[bundle] Renamed exposed services' name to classes' FQCN

Lctrs:


makasim:
doesn't it contradicts to https://symfony.com/doc/current/service_container/autowiring.html#public-and-reusable-bundles?

Lctrs:
No. It means that bundles should configure themselves their internal and exposed services. Here, those new aliases are meant to allow users of Enqueue to autowire Enqueue's bundle's services into their own classes in a Symfony app.

makasim:
So the bundle may contain autowiring declarations but it shouldn't use them itself, right?

Lctrs:
Yep. Anyway, by declaring our services with their arguments in a config file as it is done actually, this bundle don't use autowiring actually.

But by not exposing the public services by their FQCN, they could not be autowired by Symfony 4 in users' services.
Currently, we are getting deprecation messages like this in a Symfony 3 app : 
`Autowiring services based on the types they implement is deprecated since Symfony 3.3 and won't be supported in version 4.0. You should rename (or alias) the "enqueue.client.producer" service to "Enqueue\Client\Producer" instead.`

Lctrs:
An alternative could be to deprecate all `enqueue.client.producer`-like service names and alias and promote FQCN as service name instead.

Lctrs:
I was thinking as something like this : 

```yaml
services:
    Enqueue\Client\Producer:
        class: 'Enqueue\Client\Producer'
        arguments:
            - '@enqueue.client.driver'
            - '@enqueue.client.rpc_factory'
            - '@enqueue.client.extensions'

    enqueue.client.producer:
        deprecated: The "%service_id%" service is deprecated since 0.9 and will be removed in 1.0. Use "Enqueue\Client\Producer" instead.
        alias: Enqueue\Client\Producer
```

But it looks like it's not supported by Symfony sadly (https://github.com/symfony/symfony/issues/24507).

makasim:
> I was thinking as something like this :

looks good. let's do it. 

The message should be: 
`The "%service_id%" service is deprecated since 0.9 and will be removed in 0.10. Use "Enqueue\Client\Producer" instead.`

makasim:
We could go without depreciation notice. I am fine with it. Leave there a yaml comment there.

Lctrs:
@makasim If tests are green, it's ready to be reviewed.

makasim:
@Lctrs still red

Lctrs:
Yep saw it.

Lowercased services identifiers in Symfony < 3.3 is annoying. I don't have time today to work on it. I'll try to fix it by monday.

Lctrs:
@makasim Failures are unrelated.

PR #247, opened 2017-10-28
[bundle] streamline profiler view when no messages were sent

### LINKED ISSUE #243: [Symfony][Minor] profiler view when no messages collected during the request
This is **really** super-minor, but no reason not to fix it as the fix is super-minor as well. :)

In Symfony profiler, Enqueue will output a header-only table if no messages were sent during the request. What all other profiler views do in this situation is something like this:

```html
<div class="empty">
    <p>No queries were performed.</p>
</div>
```

dkarlovi:
Closes #243.


dkarlovi:
CI failures are unrelated.

Issue #213, opened 2017-10-03
[Symfony Bundle] Queue Symfony Commands (question)


jverdeyen:
Are there any future plans for an integration of a queue for console command execution?
To run and schedule Symfony console commands as background jobs with enqueue.

Kind of functionality that [JMSJobQueueBundle](https://github.com/schmittjoh/JMSJobQueueBundle) has.

I would like to integrate this into the enqueue Symfony Bundle or create a separate extension bundle for this.

👍🏼 Thanks!

makasim:
I am going to write a blog post that explains how to do it. 

For now, I can only suggest you look at code snippets that I've prepared for the blog post:

* https://github.com/php-enqueue/enqueue-sandbox/blob/cf3fc1bede66bab3bb32702ca9d044cd52db14a5/symfony/src/AppBundle/Controller/DefaultController.php#L134
* https://github.com/php-enqueue/enqueue-sandbox/blob/cf3fc1bede66bab3bb32702ca9d044cd52db14a5/symfony/src/AppBundle/Async/RunCommandProcessor.php#L12

makasim:
It is worth looking at `enqueue/job-queue`. It provides similar features plus some more advanced like sub jobs handling and exclusive job guaranty

jverdeyen:
Thanks! Very helpful!

makasim:
feel free to re-open if there is something left.

jverdeyen:
We could develop some defaults in the bundle? A job producer with an easy/basic interface and a default processor?

makasim:
@jverdeyen  I don't mind

Xymanek:
I would like to switch away from jms job due to very stale development (eg. [this](https://github.com/schmittjoh/JMSJobQueueBundle/issues/178) and [this](https://github.com/schmittjoh/JMSJobQueueBundle/issues/182)). This feature would really nice

makasim:
@Xymanek for this issue https://github.com/schmittjoh/JMSJobQueueBundle/issues/178 we have this extension https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue-bundle/Consumption/Extension/DoctrinePingConnectionExtension.php. You just have to enable it in the config.

Xymanek:
What I meant is that it wasn't merged there even though I made a PR - that's why I want to switch to this library. But lack of standardized/built-in support for queuing symfony commands is a deal-breaker for me - the app is built around it...

[...]

PR #250, opened 2017-11-02
[WIP] Add Command processor and Command creator

jverdeyen:
First steps to integrate something like the `JMSJobQueueBundle` in Enqueue.
This is a WIP and RFC.

Related Issue: https://github.com/php-enqueue/enqueue-dev/issues/213

### DIFF
--- a/pkg/enqueue/Symfony/Creator/JobCommandCreator.php
+++ b/pkg/enqueue/Symfony/Creator/JobCommandCreator.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace Enqueue\Symfony\Creator;
+
+use Enqueue\Client\ProducerInterface;
+
+class JobCommandCreator
+{
+    /**
+     * @var ProducerInterface
+     */
+    protected $producer;
+
+    /**
+     * @var string
+     */
+    protected $env;
+
+    /**
+     * CommandJobCreator constructor.
+     *
+     * @param ProducerInterface $producer
+     * @param string            $env
+     */
+    public function __construct(ProducerInterface $producer, $env)
+    {
+        $this->producer = $producer;
+        $this->env = $env;
+    }
+
+    /**
+     * @param $command
+     * @param mixed $args
+     *
+     * @return \Enqueue\Rpc\Promise|null
+     */
+    public function scheduleCommand($command, $args = [])
+    {
+        $argumentString = $this->createArgumentString($args);
+
+        return $this->producer->sendCommand('run_command', sprintf('%s %s', $command, $argumentString));
+    }
+
+    /**
+     * @param array $arguments
+     *
+     * @return string
+     */
+    public function createArgumentString(array $arguments)
+    {
+        $optionList = [];
+
+        foreach ($arguments as $key => $value) {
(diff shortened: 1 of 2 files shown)


pkg/enqueue/Symfony/Creator/JobCommandCreator.php
```diff
@@ -0,0 +1,66 @@
...
+     * @param array $arguments
+     *
+     * @return string
+     */
+    public function createArgumentString(array $arguments)
+    {
+        $optionList = [];
+
+        foreach ($arguments as $key => $value) {
+            if (!is_int($key)) {
```
makasim: what is the purpose of this check? why int?

pkg/enqueue/Symfony/Processor/RunCommandProcessor.php
```diff
@@ -0,0 +1,44 @@
...
+    public function __construct($projectDir)
+    {
+        $this->projectDir = $projectDir;
+    }
+
+    public function process(PsrMessage $message, PsrContext $context)
+    {
+        $commandline = $message->getBody();
+
+        $process = new Process('./bin/console '.$commandline, $this->projectDir);
```
makasim: how about using PhpExecutableFinder here?

pkg/enqueue/Symfony/Processor/RunCommandProcessor.php
```diff
@@ -0,0 +1,44 @@
...
+
+        try {
+            $process->mustRun();
+
+            return Result::ACK;
```
makasim: should process send a reply back to the caller?  It could be simply a command exit status or event output.

pkg/enqueue/Symfony/Processor/RunCommandProcessor.php
```diff
@@ -0,0 +1,44 @@
...
+        } catch (ProcessFailedException $e) {
+            return Result::reject(sprintf('The process failed with exception: "%s" in %s at %s', $e->getMessage(), $e->getFile(), $e->getLine()));
+        }
+    }
+
+    public static function getSubscribedCommand()
+    {
+        return 'run_command';
```
makasim: Would be good to have constant for this value. It is better to rename it to `symfony.run_command`

pkg/enqueue/Symfony/Creator/JobCommandCreator.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Enqueue\Symfony\Creator;
+
+use Enqueue\Client\ProducerInterface;
+
+class JobCommandCreator
```
makasim: Instead of a new layer over producer I would add a RunCommandMessage which accept command name, arguments, options on its constructor, implmenets JsonSerializer interface

```php
<?php

$message = new RunCommandMessage('debug:container', [], ['--env' => 'dev']);

$producer->send(RunCommandProcessor::COMMAND, $message);
```

wdyt?

makasim:
close in favor https://github.com/php-enqueue/enqueue-dev/pull/403

PR #256, opened 2017-11-05
[rdkafka] Topics created by context should contain default topic config

tPl0ch:
Currently the topics created via `RdKakfaContext::createTopic` do not contain the default topic configuration. This is unexpected, since the context knows about the topic config and should pass this information on to created topics.

### DIFF
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -25,6 +25,11 @@ class RdKafkaContext implements PsrContext
     private $conf;
 
     /**
+     * @var TopicConf
+     */
+    private $defaultTopicConf;
+
+    /**
      * @var Producer
      */
     private $producer;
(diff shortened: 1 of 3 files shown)


tPl0ch:
Hmm, the test failures are weird. Running the `Enqueue\RdKafka\Tests\RdKafkaContextTest::testTopicCreatedByContextShouldContainDefaultTopicConf` case locally works fine. Not sure why it fails with Travis.

makasim:
I don`t think there is a need to set a default topic conf to a topic object.  

According to [the doc](https://arnaud-lb.github.io/php-rdkafka/phpdoc/rdkafka.newtopic.html): the topic conf is optional and if it is not set the server should use the default one. 

```
public RdKafka\Topic RdKafka::newTopic ( string $topic_name [, RdKafka\TopicConf $topic_conf = NULL ] )

topic_conf
Optional configuration that will be used instead of the default topic configuration.
```

I spotted one potential bug [here](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/RdKafkaProducer.php#L47) that might help you out. The `$destination->getConf()` must return null of any custom options were set.

makasim:
Feel free to reopen if https://github.com/php-enqueue/enqueue-dev/pull/263 hasn't solved the issue.

Issue #255, opened 2017-11-05
[rdkafka] Introduce KeySerializer 


makasim:
It will be a better approach for https://github.com/php-enqueue/enqueue-dev/pull/254

makasim:
PR is open. all discussion is there

PR #257, opened 2017-11-06
[WIP][rdkafka] Introduce KeySerializer

tPl0ch:
This is a WIP prototype to introduce the concept of key serialization as discussed in https://github.com/php-enqueue/enqueue-dev/issues/255 and https://github.com/php-enqueue/enqueue-dev/pull/254.

The following steps will be taken in order to implement the feature:

- [x]  Define `KeySerializer` interface
- [x] Implement default `NoOpKeySerializer` incl. tests
- [ ] Implement `JsonKeySerializer` incl. tests
- [x] Add `KeySerializerAwareTrait`
- [ ] Integrate key serialization into `RdKafkaProducer` incl. tests
- [ ] Integrate key de-serialization into `RdKafkaConsumer` incl. tests
- [x] Extract `json_*` functions

### DIFF
--- a/pkg/rdkafka/NoOpKeySerializer.php
+++ b/pkg/rdkafka/NoOpKeySerializer.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Enqueue\RdKafka;
+
+class NoOpKeySerializer implements KeySerializer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function toString($key)
+    {
+        return $key;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function toKey($serialized)
+    {
+        return $serialized;
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,7 @@
   "autoload": {
     "files": [
       "pkg/enqueue/functions_include.php",
+      "pkg/rdkafka/functions.php",
       "pkg/rdkafka/Tests/bootstrap.php"
     ],
     "psr-0": {
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
   "autoload": {
     "files": [
       "pkg/enqueue/functions_include.php",
-      "pkg/rdkafka/functions.php",
+      "pkg/rdkafka/functions_include.php",
       "pkg/rdkafka/Tests/bootstrap.php"
     ],
     "psr-0": {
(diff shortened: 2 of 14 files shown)


makasim:
It has just occurred to me that the key could be serialized before setting it to the message (outside of the producer). Would it be better and simpler to go that way?

tPl0ch:
I would be confused if the serialization code for messages and keys was in different parts. You could also use the same argument for moving the serialization of  the body/value of the message. Also it might be that you want to have different Serializers for different Producers.

pkg/rdkafka/NoOpKeySerializer.php
```diff
@@ -0,0 +1,22 @@
...
+namespace Enqueue\RdKafka;
+
+class NoOpKeySerializer implements KeySerializer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function toString($key)
+    {
+        return $key;
```
makasim: maybe we should force casting to a string here

tPl0ch: That would not be a `NoOp` anymore IMO, since casting would actually change they type. I would rather just keep the current behaviour and just pass the key through by default. What do you think?

makasim: Than it contradicts to the interface contract. The KeySerializer toString methods returns string and I as a client should be able to rely on it. 

Also the rdkafka producer topic declares a key as a string 

https://arnaud-lb.github.io/php-rdkafka/phpdoc/class.rdkafka-producertopic.html

tPl0ch: OK, I agree. But then I'd rather rename it to `DefaultSerializer` to make it clear that there might be a type conversion.

makasim: > OK, I agree. But then I'd rather rename it to DefaultSerializer to make it clear that there might be a type conversion.

Sounds good.

tPl0ch:
@makasim I just had a different Idea actually. How about using a `SerializingMessageDecorator`? This way we could actually reduce the complexity in the Producer/Consumer classes and have a clean SRP inside the Decorator. By default, we could create a Decorator with `JsonSerializer` and `DefaultKeySerializer`.

makasim:
@tPl0ch one class that implements two interfaces?

PR #258, opened 2017-11-09
[Redis] add custom database index

IndraGunawan:
add configuration to use custom redis database index

reference
https://github.com/nrk/predis/wiki/Connection-Parameters
https://github.com/phpredis/phpredis/blob/develop/README.markdown#select


makasim:
@IndraGunawan Is the feature ready?

IndraGunawan:
yes, the database selection support both predis and phpredis

Issue #259, opened 2017-11-09
On exception Context is not set 


pavkatar:
On exception thrown from one of the consumers, the Context, which is sent to all extensions is not populated with a message, queue, consumer, etc. 

```php
$context = new Context($this->psrContext);
$context->setLogger($logger);
$context->setPsrQueue($consumer->getQueue());
$context->setPsrConsumer($consumer);
$context->setPsrProcessor($currentProcessor);
$context->setPsrMessage($message);
```

This is code is in the callback, but when the callback exit, the context sent to $extension->onInterrupted() is empty. QueueConsumer.php (line: 240+)

makasim:
the message is not set if [this](https://github.com/php-enqueue/enqueue-dev/blob/42443eeacba4408e0333969887f1978e0990834e/pkg/enqueue/Consumption/QueueConsumer.php#L274) condition is true.

it is definitely a bug

PR #264, opened 2017-11-10
[consumption] fix context is missing message on exception.

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/259


makasim:
@pavkatar could you test it?

https://github.com/formapro-forks/enqueue/tree/consumption-message-not-set-to-context

pavkatar:
It works 👍 
```
2017-11-11T13:46:23+00:00 INFO (6): [StatusManagerExtension] Message sent to  queue "status" with status: "error"
```

PR #269, opened 2017-11-15
Add Symfony4 support

makasim:
[ref](http://symfony.com/blog/helping-prepare-for-symfony-4-bundle-support)


makasim:
blocker https://github.com/php-http/client-common/pull/88

Issue #279, opened 2017-11-30
[dbal] Store id (uuid) as binary data.


makasim:
It turned out that Doctrine [uses string type](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html) internally for most of the drivers (except Postgres) 

As written [here](https://www.stitcher.io/blog/optimised-uuids-in-mysql) it is much better to store UUIDs as binary data. 

I propose changing id column to be binary and adding a new with a string representation (for a human).

makasim:
http://mysqlserverteam.com/storing-uuid-values-in-mysql-tables/

makasim:
PR is opened

Issue #276, opened 2017-11-26
[enqueue/dbal] Logic for "The platform does not support UUIDs natively" is incorrect


msheakoski:
In the file: https://github.com/php-enqueue/enqueue-dev/blob/332542467a10a884145f324671413478a6d022fb/pkg/dbal/DbalContext.php#L173-L175 the test should be negated with a `!`. Currently, the exception is thrown if a database DOES support UUIDs.

PR #280, opened 2017-11-30
[0.9][BC break][dbal] Store UUIDs as binary data. Improves performance

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/279
fixes https://github.com/php-enqueue/enqueue-dev/issues/276

### DIFF
--- a/pkg/dbal/DbalProducer.php
+++ b/pkg/dbal/DbalProducer.php
@@ -10,7 +10,9 @@ use Interop\Queue\InvalidMessageException;
 use Interop\Queue\PsrDestination;
 use Interop\Queue\PsrMessage;
 use Interop\Queue\PsrProducer;
+use Ramsey\Uuid\Codec\OrderedTimeCodec;
 use Ramsey\Uuid\Uuid;
+use Ramsey\Uuid\UuidFactory;
 
 class DbalProducer implements PsrProducer
 {
@@ -35,11 +37,17 @@ class DbalProducer implements PsrProducer
     private $context;
 
     /**
+     * @var OrderedTimeCodec
+     */
+    private $uuidCodec;
+
+    /**
      * @param DbalContext $context
      */
     public function __construct(DbalContext $context)
     {
         $this->context = $context;
+        $this->uuidCodec = new OrderedTimeCodec((new UuidFactory())->getUuidBuilder());
     }
 
     /**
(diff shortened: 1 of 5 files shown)


dkarlovi:
` ramsey/uuid-doctrine` provides a [time-ordered binary UUID storage](https://github.com/ramsey/uuid-doctrine#innodb-optimised-binary-uuids).

makasim:
@dkarlovi unfortunately it requires `doctrine/orm` which is not an option for us. 
The transport should work with doctrine/dbal.

pkg/dbal/DbalProducer.php
```diff
@@ -74,15 +75,12 @@ public function send(PsrDestination $destination, PsrMessage $message)
...
         }
 
-        $sql = 'SELECT '.$this->context->getDbalConnection()->getDatabasePlatform()->getGuidExpression();
-        $uuid = $this->context->getDbalConnection()->query($sql)->fetchColumn(0);
-
-        if (empty($uuid)) {
-            throw new \LogicException('The generated uuid is empty');
-        }
+        $hasNativeGuid = $this->context->getDbalConnection()->getDatabasePlatform()->hasNativeGuidType();
+        $uuid = Uuid::uuid4();
```
msheakoski: Why not `Uuid::uuid1()`? It is more efficient for the database to index than a random v4 UUID.

makasim: Honestly, I dont know I've been using uuid4 everywhere. I'll look into it. 

Thanks.

makasim: good write up on this  https://www.sohamkamani.com/blog/2016/10/05/uuid1-vs-uuid4/

dkarlovi: @makasim check the link from Percona linked in my comment.

makasim: @dkarlovi yeah, I saw it. Unfortunately I am not sure we can adopt it since it is MySQL\InnoDB  specific stuff. We have to support or DBAL transport hence the solution should be flexible enough to work on all of those platforms

dkarlovi: No, I meant the influence of random on performance. v1 vs v4 should be considered here IMO.

makasim: ah, I see

makasim:
@dkarlovi updated PR with uuid1 and the use of time ordered codec.

makasim:
Publish and consume 100 000 messages

enqueue/dbal 0.8

Publish took 80.730087995529 seconds, 40.152kb memory
Consume took 135.10051894188 seconds, 196.096kb memory

enqueue/dbal 0.9 

Publish took 74.140038013458 seconds, 191.248kb memory
Consume took 149.82619380951 seconds, 137.888kb memory

It turned out that the difference is not that big.

Issue #290, opened 2017-12-14
Consumer Requeue -> DBAL NotNullConstraintViolationException


SchumacherFM:
Hi there,

Thanks for this great queuing framework!

It runs in Magento2 very smoothly but I have to use DBAL backend which brings an error:

The automatically created `enqueue` DB table structure looks:
```
CREATE TABLE `enqueue` (
  `id` char(36) COLLATE utf8_unicode_ci NOT NULL COMMENT '(DC2Type:guid)',
  `published_at` bigint(20) NOT NULL,
  `body` longtext COLLATE utf8_unicode_ci,
  `headers` longtext COLLATE utf8_unicode_ci,
  `properties` longtext COLLATE utf8_unicode_ci,
  `redelivered` tinyint(1) DEFAULT NULL,
  `queue` varchar(255) COLLATE utf8_unicode_ci NOT NULL,
  `priority` smallint(6) NOT NULL,
  `delayed_until` int(11) DEFAULT NULL,
  `time_to_live` int(11) DEFAULT NULL,
  PRIMARY KEY (`id`),
  KEY `IDX_CFC35A68E0D4FDE1` (`published_at`),
  KEY `IDX_CFC35A687FFD7F63` (`queue`),
  KEY `IDX_CFC35A6862A6DC27` (`priority`),
  KEY `IDX_CFC35A681A065DF8` (`delayed_until`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```

When a consumer now requeues then the following error will occur:

```
[error] Consuming interrupted by exception

  [Doctrine\DBAL\Exception\NotNullConstraintViolationException]
  An exception occurred while executing 'INSERT INTO enqueue (body, headers, properties, priority, queue, redelivered) VALUES (?, ?, ?, ?, ?, ?)' with params ["msg", "{\"content_type\":\"application\\\/json\",\
  "message_id\":\"c0165f40-d05c-44f4-a3ae-2d8286a89b86\",\"timestamp\":1513248090,\"reply_to\":null,\"correlation_id\":null}", "{\"enqueue.topic_name\":\"xxx.vote.new\",\"enqueu
  e.processor_name\":\"xxx-votes\",\"enqueue.processor_queue_name\":\"default\"}", 2, "enqueue.app.default", 1]:

  SQLSTATE[HY000]: General error: 1364 Field 'id' doesn't have a default value

[...]

PR #291, opened 2017-12-14
[dbal] Fix message re-queuing. Reuse producer for it.

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/290


SchumacherFM:
Thanks!

PR #297, opened 2017-12-19
Fix reversed logic for native UUID detection

msheakoski:
I know that you are working on #280 but for current users of Enqueue, it is very inconvenient to manually edit the package to fix this bug and it also makes deploying/CI difficult. Can you release this as 0.8.12?

### DIFF
--- a/pkg/dbal/DbalContext.php
+++ b/pkg/dbal/DbalContext.php
@@ -170,7 +170,7 @@ class DbalContext implements PsrContext
             return;
         }
 
-        if ($this->getDbalConnection()->getDatabasePlatform()->hasNativeGuidType()) {
+        if (!$this->getDbalConnection()->getDatabasePlatform()->hasNativeGuidType()) {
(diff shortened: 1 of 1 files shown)


pkg/dbal/DbalContext.php
```diff
@@ -170,7 +170,7 @@ public function createDataBaseTable()
             return;
         }
 
-        if ($this->getDbalConnection()->getDatabasePlatform()->hasNativeGuidType()) {
+        if (!$this->getDbalConnection()->getDatabasePlatform()->hasNativeGuidType()) {
```
makasim: I guess this change would break MySql. Could you remove it?

msheakoski: I have removed the check. That check should not be necessary in the first place because Doctrine will use the UUID column type if available, otherwise fall back to VARCHAR. One thing to take note of is that it may be necessary to add some code to #280 to detect and migrate the column type if you decide to merge that PR.

PR #299, opened 2017-12-20
Check if logger exists

pascaldevink:
When handling a signal, a logger is used to log the catched signals. However, the logger can sometimes be uninitialised, which resulted in errors.


makasim:
@pascaldevink looks great! Could you please fix [CS issue](https://travis-ci.org/php-enqueue/enqueue-dev/jobs/319195723) so I can merge it?

pascaldevink:
@makasim I've fixed the CS issue, but I'm pretty sure the failing tests are not caused by these changes. Any chance master is also broken?

makasim:
It is a known issue, there are some env vars that store Amazon SQS credentials, and Travis does not add them to the build if it initiated by outside contributor. Hence they fail, and honestly I don't know what to do with it

Sure I'll merge it.

PR #315, opened 2018-01-08
[Sqs] Allow array-based DSN configuration

beryllium:
I encountered a problem with using DSN-based configuration in Symfony 4, where the 'dsn' value was always being passed as an array key. The constructor did not properly handle this, and the primary symptom of this was a constant complaint that the 'region' was not set, even though it was present in the DSN.

### DIFF
--- a/pkg/sqs/SqsConnectionFactory.php
+++ b/pkg/sqs/SqsConnectionFactory.php
@@ -39,11 +39,18 @@ class SqsConnectionFactory implements PsrConnectionFactory
     {
         if (empty($config) || 'sqs:' === $config) {
             $config = [];
-        } elseif (is_string($config)) {
+        }
+
+        if (is_string($config)) {
             $config = $this->parseDsn($config);
-        } elseif (is_array($config)) {
-        } else {
-            throw new \LogicException('The config must be either an array of options, a DSN string or null');
+        }
+
+        if (!is_array($config)) {
+            throw new \LogicException('The config must be either an array of options, a DSN string, or null');
+        }
+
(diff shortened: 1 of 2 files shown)


pkg/sqs/SqsConnectionFactory.php
```diff
@@ -39,11 +39,18 @@ public function __construct($config = 'sqs:')
...
-        } elseif (is_array($config)) {
-        } else {
-            throw new \LogicException('The config must be either an array of options, a DSN string or null');
+        }
+
+        if (!is_array($config)) {
+            throw new \LogicException('The config must be either an array of options, a DSN string, or null');
+        }
+
+        if (count($config) == 1 && array_key_exists('dsn', $config)) {
```
makasim: dsn option could be used with other options in config. It should work like [here](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-tools/ConnectionConfig.php#L137)
  
The config is treat as default options, and options from dsn could overwrite them

beryllium:
Looks like the tests are failing due to a misconfiguration with IAM credentials? Unless I'm mistaken, it doesn't seem to be related to the PR.

makasim:
It is a security policy, Travis does not add secret envs vars if PR author is not member of org.

makasim:
that should fix it https://github.com/php-enqueue/enqueue-dev/commit/dfd44991f94c46f3ded25bb3c6805ad13964109e

beryllium:
Merged the latest master, hopefully everything will pass now :)

makasim:
Thank you

PR #319, opened 2018-01-10
Update quick_tour.md

errogaht:


makasim:
This syntax is valid and is supposed to work (there are tests for that, for example [this one](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-tools/Tests/ConnectionConfigTest.php#L182)). It is used in [real life application](https://github.com/php-comrade/comrade-dev/blob/96f26342bf5e41b4bb3e48c175ce8c7b6f6bfa06/apps/jm/config/packages/enqueue.yaml#L3) as expected.

Why do you think it does not work? Do you see any errors or exceptions?

makasim:
As far as I know `schema:` syntax is a valid URI accoding to [RFC 3986](https://tools.ietf.org/html/rfc3986)

makasim:
`amqp:` is a short equivalent of `amqp://guest:guest@localhost:5672/%2f`

PR #326, opened 2018-01-11
Update config_reference.md

errogaht:
it is working for me with this configuration. I use "enqueue/enqueue-bundle": 0.8.13.
with your version when I run bin/console enqueue:consume default --setup-broker i receiving Processor was not found. processorName: "enqueue.client.router_processor"

### DIFF
--- a/docs/bundle/config_reference.md
+++ b/docs/bundle/config_reference.md
@@ -251,7 +251,7 @@ enqueue:
         app_name:             app
         router_topic:         default
         router_queue:         default
-        router_processor:     enqueue.client.router_processor
+        router_processor:     Enqueue\Client\RouterProcessor
         default_processor_queue: default
         redelivered_delay_time: 0
     consumption:


makasim:
It works for me, it works in tests and most importantly the doc contains an output of `./bin/console config:dump-reference enqueue` and should not be edited manually.

errogaht:
hmm. why it's not working for me? I have symfony 2.8. and when client: ~ all work fine, but when I put parameters block with client from documentation  got "Processor was not found. processorName: "enqueue.client.router_processor"" then if I change router_processor: to Enqueue\Client\RouterProcessor all work fine. now all work fine for me, maybe it can useful for you guys, I don't need any help with this because I fix it

makasim:
the doc is [out of sync with the code](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue-bundle/DependencyInjection/Configuration.php#L50), your fix is correct.

makasim:
Thank you

PR #339, opened 2018-01-19
[client] Add ability to preapre message, or use no JSON format.

makasim:


Koc:
What about providing serialization layer instead of extensions? 

Create simple interface like this https://github.com/php-enqueue/enqueue-dev/blob/2b1094c3ff6db885cc999ce15291f598f352da6a/pkg/rdkafka/Serializer.php and ship lib with 2 serializers: json (simple, native like now) and Symfony's serializer.

makasim:
It would introduce one extra concept, and it would be possible to achieve the goal with extensions. There might be an extension that supports Symfony serializer.

makasim:
closed in favor https://github.com/php-enqueue/enqueue-dev/pull/517

Issue #340, opened 2018-01-21
[dbal] Consumer never fetches messages ordered by published time


f7h:
The database column "priority" is defined as NOT NULL if you use the DbalContext class to generate the schema.

The default priority of a DbalMessage is 0.

The fetchPrioritizedMessage() method from the DbalConsumer class fetches messages where the priority IS NOT NULL, so all of them.

Consequently, fetchMessage(), which fetches the oldest message first, is never called as long as there are any messages left in the queue.

The priority column should be nullable and the default priority should also be null in my opinion.

makasim:
@f7h would you like to send us a PR?

PR #343, opened 2018-01-22
[dbal] Consumer never fetches messages ordered by published time

f7h:
#340


makasim:
I am going to merge this to 0.9 branch (since it contains BC breaks).

makasim:
I've sync this repo with read-only sub tree splits, so to make it easy for you to use `enqueue/gearman` with the fix. 

Though I'd recommend setting the commit hash as a version in your composer.json. There are BC breaks expected in 0.9 version and it is not stable

f7h:
Thank you.

Issue #270, opened 2017-11-17
gearmand queue library can't not use for php7


eeeeeeeecccccccc:
Using version ^0.8.0 for enqueue/gearman
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for enqueue/gearman ^0.8.0 -> satisfiable by enqueue/gearman[0.8.0].
    - enqueue/gearman 0.8.0 requires ext-gearman ^1.1 -> the requested PHP extension gearman has the wrong version (2.0.3) installed.

https://blog.remirepo.net/pages/PECL-extensions-RPM-status

  | gearman | 1.1.2 (php 5, stable)2.0.3 (php 7) | PHP wrapper to libgearmanv1 is from pecl, v2 is from wcgallego fork. | 5.5, 5.5, 5.6, 7.0, 7.1, 7.2
-- | -- | -- | -- | --

makasim:
Are the extension APIs the same? If so we can just update the composer req.

Could you test it, please?

Jawshua:
Hey guys,

We've been testing PHP 7.2 with the Gearman 2.0.3 extension and haven't had to change any of our gearman related code, so it _should_ be safe to switch the constraint to `^1.1 || ^2.0`.

I ran the tests on enqueue/gearman 0.8.12 using PHP 7.2.1 (CentOS 7.4, php72 collection from https://blog.remirepo.net):
```
gearman # export GEARMAN_DSN=gearman://127.0.0.1:4730                                                                                                                                                                                            
gearman # vendor/bin/phpunit 
PHPUnit 4.8.34-1167-g53951b4 by Sebastian Bergmann and contributors.

............................................................      60 / 60 (100%)

Time: 5.56 seconds, Memory: 4.00MB
```

makasim:
Great! Could you please send a PR (use 0.9 branch as base one)?

PR #347, opened 2018-01-24
Gearman PHP 7 support

Jawshua:
Fixes https://github.com/php-enqueue/enqueue-dev/issues/270


makasim:
I've sync this repo with read-only sub tree splits, so to make it easy for you to use `enqueue/gearman` with the fix. 

Though I'd recommend setting the commit hash as a version in your composer.json. There are BC breaks expected in 0.9 version and it is not stable

PR #350, opened 2018-02-01
[redis] Authentication support added

### LINKED ISSUE #349: Authentication Support for Redis
The redis transport doesn't support authentication.
But this required to use redis in environments like Heroku/AWS.

FrankGiesecke:
Use the authentication information provided by the config after parsing a dsn.

Closes #349

### DIFF
--- a/pkg/redis/PhpRedis.php
+++ b/pkg/redis/PhpRedis.php
@@ -82,6 +82,14 @@ class PhpRedis implements Redis
                 );
             }
 
+            if (array_key_exists('pass', $this->config)) {
+                $this->config['auth'] = $this->config['pass'];
+            }
+
+            if (array_key_exists('auth', $this->config)) {
+                $this->redis->auth($this->config['auth']);
+            }
+
             $this->redis->select($this->config['database']);
         }
 


pkg/redis/PhpRedis.php
```diff
@@ -82,6 +82,14 @@ public function connect()
...
             }
 
+            if (array_key_exists('pass', $this->config)) {
+                $this->config['auth'] = $this->config['pass'];
+            }
+
+            if (array_key_exists('auth', $this->config)) {
+                $this->redis->auth($this->config['auth']);
+            }
+
```
makasim: not sure we need the first if `if (array_key_exists('pass', $this->config)) {`.

```php
<?php
            if (array_key_exists('pass', $this->config)) {
                $this->redis->auth($this->config['pass']);
            }
```

FrankGiesecke: This is needed for services like Herold.
They provide a token for authentication in the password part.

makasim:
This will be possible once #372 is merged. Like this:

```php
<?php
use Enqueue\Redis\RedisConnectionFactory;
use Enqueue\Redis\PRedis;
 
$config = []; 
$options = [];

$predis = new \PRedis\Client($config, $options);
$predis->auth('aPassword');
$redis = new PRedis();

$factory = new RedisConnectionFactory(['vendor' => 'custom', 'redis' => $redis]);
```

makasim:
Yes, the Enqueue\PhpRedis constructor has to be adjusted to accept the instance of `\Redis` class.

FrankGiesecke:
Please reconsider your decision. It would be very useful to consume the authentication from within the DSN. Creating an own instance of a redis client only to support authentication is much more overhead and not necessary.

MaSpeng:
I completely agree, when using various large hosters an authorization to the redis Instance is required. Only for this purpose a pre-configured redis instance must be explicitly created instead of drawing the maximum benefit from the existing functionalities, especially since the adaptation effort is apparently very low and non-invasive.

dgoujard:
I agree too, it's not so uncommon to use redis with password, it's not really practical to create an custom vendor. It's more easy with DSN, other transporters handle password mechanism in DSN

FrankGiesecke:
@makasim I'm glad to hear this and I would be appreciated if you wold merge my changes.

pkg/redis/PhpRedis.php
makasim: The code I posted does exactly the same but with one `if` statement.

makasim:
Can you add auth support to predis too?

makasim:
and some tests.

makasim:
add auth option to the doc block https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/redis/RedisConnectionFactory.php#L31

makasim:
merged as part of https://github.com/php-enqueue/enqueue-dev/pull/497. both ways are supported 

you can pass  `pass` or an instance of `Enqueue\Redis`.

Issue #352, opened 2018-02-05
[SQS] Allow custom aws endpoint configuration


gitis:
SqsConnectionFactory does not allow to provide custom endpoint configuration even though SqsClient supports it. Therefore there is no way to test other than aws implementations.

Ability to configure endpoint would provide the ability to consume messages locally using localstack or other similar services.

makasim:
PR is very welcome.

gitis:
it's open https://github.com/php-enqueue/enqueue-dev/pull/353

PR #353, opened 2018-02-05
Added endpoint configuration and updated the tests

gitis:
https://github.com/php-enqueue/enqueue-dev/issues/352


makasim:
Thank you.

PR #354, opened 2018-02-05
Automatically enable traceable producer depends on debug

Koc:
Make same behavior like in all other bundles.

TODO:
- [ ] fix tests

### DIFF
--- a/pkg/enqueue-bundle/DependencyInjection/Configuration.php
+++ b/pkg/enqueue-bundle/DependencyInjection/Configuration.php
@@ -45,7 +45,7 @@ class Configuration implements ConfigurationInterface
         }
 
         $rootNode->children()
-            ->arrayNode('client')->addDefaultsIfNotSet()->children()
+            ->arrayNode('client')->children()
                 ->booleanNode('traceable_producer')->defaultValue($this->debug)->end()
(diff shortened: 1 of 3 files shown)


makasim:
wouldn't it be better to do it on an app or a recipe level like this?

```yml
enqueue
   client: 
     traceable_producer: %kernel.debug%
```

Koc:
you can do this also or use any other parameter what you want. But for zero-line configuration better using suggested approach. It widely used already in many bundles, like: [DoctrineBundle](https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Configuration.php#L25), [SncRedisBundle](https://github.com/snc/SncRedisBundle/blob/master/DependencyInjection/Configuration/Configuration.php#L27) and many other

Koc:
Even more, this approach suggested by Symfony: http://symfony.com/doc/current/configuration/using_parameters_in_dic.html

Only one thing: did you know why `testShouldConfigureFooTransport` fails?

makasim:
@Koc ok let's do it then.  I don't know why it fails.

makasim:
merged

Koc:
@makasim thank you. Have you fixed tests? I've tryed but cann't

makasim:
@Koc sure, they are fixed

PR #359, opened 2018-02-07
Do not cast microtime to integer

Gugiman:
The benefit of precise scheduling using microtime is gone when casting microtime output to integer

var_dump((int) microtime(true) * 10000);
// int(15180039750000)

var_dump(time());
// int(1518003975)

var_dump(microtime(true) * 10000);
// float(15180039759562)

### DIFF
--- a/pkg/dbal/DbalProducer.php
+++ b/pkg/dbal/DbalProducer.php
@@ -83,7 +83,7 @@ class DbalProducer implements PsrProducer
 
         $dbalMessage = [
             'id' => $uuid,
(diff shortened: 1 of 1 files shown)


makasim:
IMO It should be `(int) (microtime(true) * 10000)`

Gugiman:
Your're right
I commited the change in this PR

makasim:
@Gugiman could you please rebase against the latest master?

Gugiman:
It sais "Everything up-to-date"

makasim:
merged.

Issue #351, opened 2018-02-05
[Docs] Describe difference between command and event messages


Koc:
Please add note in the [quick tour](https://github.com/php-enqueue/enqueue-dev/blob/master/docs/quick_tour.md) what difference between command and event messages and when to use each of them. For now there is only examples how to send them, without explanation.

PR #365, opened 2018-02-09
[client][skip ci] Explain meaning of sendEvent, sendCommand methods.

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/351

### DIFF
--- a/pkg/simple-client/SimpleClient.php
+++ b/pkg/simple-client/SimpleClient.php
@@ -125,6 +125,8 @@ final class SimpleClient
     }
 
     /**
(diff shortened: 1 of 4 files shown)


pkg/simple-client/SimpleClient.php
```diff
@@ -125,6 +125,8 @@ public function sendEvent($topic, $message)
     }
 
     /**
+     * @deprecated since 0.8.18 and will be removed in 0.9. Use sendEvent method instead
```
Koc: can you trigger deprecation notice also?

Issue #367, opened 2018-02-12
Pass options to predis client when using redis transport


fvaresi:
It would be nice to be able to pass [client options](https://github.com/nrk/predis/wiki/Client-Options) to predis transport.

Example:
```
<?php
use Enqueue\Redis\RedisConnectionFactory;

$connectionFactory = new RedisConnectionFactory([
    'host' => 'localhost',
    'port' => 6379,
    'vendor' => 'predis',
    ...
    'options' => ['cluster' => 'predis', ...],
]);
```

makasim:
This will be possible once #372 is merged. Like this:

```php
<?php
use Enqueue\Redis\RedisConnectionFactory;

use Enqueue\Redis\PRedis;
 
$config = [
    'host' => 'localhost',
    'port' => 6379,
    'vendor' => 'predis',
]; 
$options = ['cluster' => 'predis'];

$redis = new PRedis(new \PRedis\Client($config, $options));

$connectionFactory = new RedisConnectionFactory([
    'redis' => $redis,
    'vendor' => 'custom',
]);
```

PR #368, opened 2018-02-12
issue #367 - pass options to predis client

fvaresi:
#367

### DIFF
--- a/pkg/redis/RedisConnectionFactory.php
+++ b/pkg/redis/RedisConnectionFactory.php
@@ -28,6 +28,7 @@ class RedisConnectionFactory implements PsrConnectionFactory
      *  'persisted' => bool, Whether it use single persisted connection or open a new one for every context
      *  'lazy' => the connection will be performed as later as possible, if the option set to true
      *  'database' => Database index to select when connected (default value: 0)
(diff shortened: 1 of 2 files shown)


makasim:
It would be better to upgrade the constructor of RedisConnectionFactory so that it accepts an instance of [Redis](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/redis/Redis.php) object.

fvaresi:
@makasim ok, will give that a try.

fvaresi:
Hi @makasim. I was reviewing the code and I think handling this as a config option is better than modifying the constructor of RedisConnectionFactory for the following reasons:

a) if we modify RedisConnectionFactory, we won't be able to continue using the RedisTransportFactory abstraction without significant modifications.
b) one could still use RedisConnectionFactory directly, but this would make it difficult to switch to a different transport in the future
c) if we use the proposed changes in config handling instead, SimpleClient would still work without any changes.

What do you think? Do you still think modifying RedisConnectionFactory is the best option?

makasim:
there is a better solution https://github.com/php-enqueue/enqueue-dev/issues/367#issuecomment-365867762

Issue #383, opened 2018-02-28
[symfony bundle] EnqueueExtension. Transport factory with such name already added. Name stomp


TheRatG:
| Q                 | A
| ----------------- | ---
| Issue Type        | Bug
| Version  | 0.8.21
| Local Machine OS  | Mac Os Darwin
| Remote Machine OS | Ubuntu

### Description 
When using ```./bin/console debug:config EnqueueBundle``` in production mode  I am getting this error

```
In EnqueueExtension.php line 48:
  Transport factory with such name already added. Name stomp  
```

### Steps to reproduce
```
composer create-project symfony/skeleton enqueue
cd enqueue
composer config extra.symfony.allow-contrib true
composer require enqueue/enqueue-bundle enqueue/amqp-lib
perl -pi -w -e 's/APP_ENV=dev/APP_ENV=prod/g;' .env
./bin/console c:cl
./bin/console debug:config EnqueueBundle
```

### Output log
```
./bin/console debug:config EnqueueBundle -vvv
2018-02-28T09:22:15+00:00 [error] Error thrown while running command "debug:config EnqueueBundle -vvv". Message: "Transport factory with such name already added. Name stomp"
2018-02-28T09:22:15+00:00 [debug] Command "debug:config EnqueueBundle -vvv" exited with code "1"

In EnqueueExtension.php line 48:
  [LogicException]                                            
  Transport factory with such name already added. Name stomp  
                                                              
Exception trace:
 Enqueue\Bundle\DependencyInjection\EnqueueExtension->addTransportFactory() at /Users/v.pak/www/enqueue/vendor/enqueue/enqueue-bundle/EnqueueBundle.php:60
```

neandher:
When this happen, just run: cache:clear and works for me.

neandher:
@DamienHarper You're right.

PR #388, opened 2018-03-01
[bundle] fix for "Transport factory with such name already added"

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/383


TheRatG:
Thanks a lot

Issue #375, opened 2018-02-19
[symfony bundle] The env "resolve:ENQUEUE_DSN" var is not defined


jralph:
When using the `resolve` ability with symfony 4's advanced env settings I am getting this error

```
[LogicException]
  The env "resolve:ENQUEUE_DSN" var is not defined
```

This is being thrown from `Enqueue\Symfony\DefaultTransportFactory` at line `160`.

This appears to only be thrown when I try to use `%env(resolve:ENQUEUE_DSN)%` in the config. Using an environment variable without resolve works, but my use case requires a dynamic variable.

Is there any fix for this and has this been noticed before? I noticed there is a `TODO: We'll have to come up with a better solution.` on the method, I'm guessing this is related. If so, is there any progress on a solution at all?

PR #393, opened 2018-03-02
[bundle] add support for %env(resolve:FOO)% feature.

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/375


thePanz:
@makasim could you rebase this MR against the latest `master`?
The solution is not working with SF 4.0.9, where the following exception is thrown:

```
Using a cast in "env(resolve:ENQUEUE_DSN)" is incompatible with resolution at compile time in
"Enqueue\Bundle\DependencyInjection\EnqueueExtension". The logic in the extension should be
moved to a compiler pass, or an env parameter with no cast should be used instead.
```

PR #396, opened 2018-03-03
Convert between Message::$expire and DbalMessage::$timeToLive

### LINKED ISSUE #391: [DbalDriver] does not convert Message::$expire to DbalMessage::$timeToLive
The following methods are missing code to handle expire/timeToLive conversion between Message and DbalMessage objects:
https://github.com/php-enqueue/enqueue-dev/blob/bc611861bf0ae458e5278b4925c32227986511ea/pkg/dbal/Client/DbalDriver.php#L61
https://github.com/php-enqueue/enqueue-dev/blob/bc611861bf0ae458e5278b4925c32227986511ea/pkg/dbal/Client/DbalDriver.php#L89
which results in the "enqueue" table having a null value for "time_to_live" even though it was set on the Message object prior to sending the event/command.

msheakoski:
Fixes a condition which results in the "enqueue" table having a null value for "time_to_live" even though it was set on the Message object prior to sending the event/command. Closes #391

### DIFF
--- a/pkg/dbal/Client/DbalDriver.php
+++ b/pkg/dbal/Client/DbalDriver.php
@@ -72,6 +72,7 @@ class DbalDriver implements DriverInterface
         $transportMessage->setMessageId($message->getMessageId());
         $transportMessage->setTimestamp($message->getTimestamp());
         $transportMessage->setDeliveryDelay($message->getDelay());
+        $transportMessage->setTimeToLive($message->getExpire());
         $transportMessage->setReplyTo($message->getReplyTo());
         $transportMessage->setCorrelationId($message->getCorrelationId());
         if (array_key_exists($message->getPriority(), self::$priorityMap)) {
@@ -97,6 +98,8 @@ class DbalDriver implements DriverInterface
         $clientMessage->setContentType($message->getHeader('content_type'));
         $clientMessage->setMessageId($message->getMessageId());
         $clientMessage->setTimestamp($message->getTimestamp());
+        $timeToLive = $message->getTimeToLive();
+        $clientMessage->setExpire((null === $timeToLive) ? null : (int) round($timeToLive));
         $clientMessage->setDelay($message->getDeliveryDelay());
         $clientMessage->setReplyTo($message->getReplyTo());
         $clientMessage->setCorrelationId($message->getCorrelationId());
--- a/pkg/dbal/Tests/Client/DbalDriverTest.php
+++ b/pkg/dbal/Tests/Client/DbalDriverTest.php
@@ -94,6 +94,7 @@ class DbalDriverTest extends \PHPUnit_Framework_TestCase
         $transportMessage->setTimestamp(1000);
         $transportMessage->setPriority(2);
         $transportMessage->setDeliveryDelay(12345);
+        $transportMessage->setTimeToLive(67890);
 
         $driver = new DbalDriver(
             $this->createPsrContextMock(),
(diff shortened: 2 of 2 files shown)


pkg/dbal/Client/DbalDriver.php
```diff
@@ -72,6 +72,7 @@ public function createTransportMessage(Message $message)
         $transportMessage->setMessageId($message->getMessageId());
         $transportMessage->setTimestamp($message->getTimestamp());
         $transportMessage->setDeliveryDelay($message->getDelay());
+        $transportMessage->setTimeToLive($message->getExpire());
```
makasim: should be `$message->getExpire() * 1000` as clients message [uses seconds](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Client/Message.php#L43) where transport [uses miliseconds](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/dbal/DbalMessage.php#L40)

pkg/dbal/Client/DbalDriver.php
```diff
@@ -97,6 +98,8 @@ public function createClientMessage(PsrMessage $message)
         $clientMessage->setContentType($message->getHeader('content_type'));
         $clientMessage->setMessageId($message->getMessageId());
         $clientMessage->setTimestamp($message->getTimestamp());
+        $timeToLive = $message->getTimeToLive();
+        $clientMessage->setExpire((null === $timeToLive) ? null : (int) round($timeToLive));
```
makasim: should be `(int) round($timeToLive / 1000)`

pkg/dbal/Tests/Client/DbalDriverTest.php
```diff
@@ -118,11 +119,29 @@ public function testShouldConvertTransportMessageToClientMessage()
...
+            $this->createDummyQueueMetaRegistry()
+        );
+
+        $transportMessage = new DbalMessage();
+
+        $transportMessage->setTimeToLive(1.4);
+        $clientMessage = $driver->createClientMessage($transportMessage);
+        $this->assertSame(1, $clientMessage->getExpire());
+
+        $transportMessage->setTimeToLive(1.5);
```
makasim: FYI. It is 1.5 milliseconds

msheakoski:
I was following the same convention used for the delivery delay in order to not break the existing behavior. Although Enqueue\Client\Message specifies that the delay is in seconds:
https://github.com/php-enqueue/enqueue-dev/blob/3a868bdd0b1d3e36d2bf1b711b2b3bec7338f689/pkg/enqueue/Client/Message.php#L198-L206
when it gets converted into a DbalMessage, it is not converted into milliseconds:
https://github.com/php-enqueue/enqueue-dev/blob/3a868bdd0b1d3e36d2bf1b711b2b3bec7338f689/pkg/dbal/Client/DbalDriver.php#L74
https://github.com/php-enqueue/enqueue-dev/blob/3a868bdd0b1d3e36d2bf1b711b2b3bec7338f689/pkg/dbal/DbalMessage.php#L204-L212
I looked through the fs and redis backends to see if they also have the same behavior, but it does not look like they support delays. How would you like to proceed?

pkg/dbal/Tests/Client/DbalDriverTest.php
msheakoski: I did this because the $transportMessage value can be `null|int|float`, but $clientMessage can only be `null|int`. Despite $clientMessage claiming to be in seconds, the current behavior of the DbalDriver is to not perform the conversion to milliseconds.

msheakoski:
Just checking if you had any thoughts on this PR and what course of action to take? I think merge it as is for 0.8 to maintain compatibility and change the seconds to milliseconds for 0.9

makasim:
The Enqueue Client sets delaying in seconds, where all queue interop transports should use milliseconds. Just make sure you properly convert seconds to milliseconds in the Driver's createClientMessage, createTransportMessage methods.

msheakoski:
Okay, in that case, the current implementation of delivery delay is broken for DBAL. If you look at my [previous comment](https://github.com/php-enqueue/enqueue-dev/pull/396#issuecomment-372160030), createTransportMessage currently does not convert from seconds to milliseconds which is the cause of this confusion. I will create a PR to fix this as well, but it will be a BC break for anybody relying on the current behavior where the delay is specified in milliseconds.

PR #401, opened 2018-03-06
[doc][magento2][skip ci] Add docs for Mangeto2 module.

makasim:


Koc:
Does it has sense store docs for module here? Some module located in separate repository, IMHO better ship docs with it.

makasim:
I make my life easier. One day all doc will be moved to its own repo.

PR #407, opened 2018-03-15
[amqp-tools] Allow for using vhost='/'

Nyholm:


makasim:
This is the right syntax for `/` as vhost: `amqp://host/%2f`. More here https://www.rabbitmq.com/uri-spec.html

Nyholm:
Okey. So if I configure my dns like ‘amqp://host/%2f’ then I would be all fine. Thanks. 

So my PR is basically just syntactic sugar to improve DX for other people like me that does not know about %2f.

makasim:
> So my PR is basically just syntactic sugar to improve DX for other people like me that does not know about %2f.

The RabbitMQ doc says that `amqp://host/` should have empty string vhost.

Nyholm:
Fair. But when I configured my DSN like `ampq://foo:bar@host:port/`, I've got a complaint that "vhost cannot be empty". (Using Bunny adapter)

makasim:
Sorry, Couldn't be merged cuz it contradicts RabbitMQ doc.

Nyholm:
I understand. Thank you for the review

PR #412, opened 2018-03-23
Fixed typos

pborreli:


makasim:
@pborreli thank for the fixes. could you please fix [CS issue](https://travis-ci.org/php-enqueue/enqueue-dev/jobs/357332820) and we good to go.

PR #414, opened 2018-03-27
[bundle] Don't ping DBAL connection if it wasn't opened

ramunasd:
Issue appears when symfony app has more than one connection defined. $connection->ping() always tries to open connection regardless it was used previously or not.

### DIFF
--- a/pkg/enqueue-bundle/Consumption/Extension/DoctrinePingConnectionExtension.php
+++ b/pkg/enqueue-bundle/Consumption/Extension/DoctrinePingConnectionExtension.php
@@ -32,6 +32,10 @@ class DoctrinePingConnectionExtension implements ExtensionInterface
     {
         /** @var Connection $connection */
         foreach ($this->registry->getConnections() as $connection) {
+            if (!$connection->isConnected()) {
+                continue;
+            }
+
             if ($connection->ping()) {
                 return;
             }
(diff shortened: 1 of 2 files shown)


makasim:
The purpose of the extension is to make sure the connection is valid, bypassing the check contradicts the goal of it. 

Just disable the extension.

ramunasd:
This extension is cool and very usefull. The problem is that it opens and keeps all connections, even those which are not used in consumer.
Let's say we have 10 dbal configs and 100 workers, then each worker tries to keep all dbal connections alive and total connection count is 1000, but only 100 is required.

makasim:
The purpose of the extension is to restore the connection which is timeouted. Is there a way to tell the difference between not used connection\new one and timeouted one?

ramunasd:
Yes, it is easy to check whether connection was opened previously or not. For this purpose DBAL\Connection contains flag $_isConnected which is accessible via ->isConnected().
So, if:
* isConnected() == true, then it means connection was opened previously and it might be lost due to some reason. Next call to ping() checks it.
* isConnected() == false, then it means connection was **not** opened previously, it is closed now and there is no need to perform any additional checks.

By default Doctrine opens connections on demand, so there is no need to ensure that all possible connections are alive.

makasim:
@ramunasd thank you for the clarification. It makes sense.

PR #417, opened 2018-03-30
Fix functional testing documentation example

sbacelic:
TracebleProducer is registered as ProducerInterface::class

### DIFF
--- a/docs/bundle/functional_testing.md
+++ b/docs/bundle/functional_testing.md
@@ -42,6 +42,7 @@ If you did so, you can use its methods `getTraces`, `getTopicTraces` or `clearTr
 ```php
 <?php
 use Symfony\Bundle\FrameworkBundle\Test\WebTestCase;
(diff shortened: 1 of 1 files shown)


makasim:
There must be a traceable producer service registered https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php#L110

If set the option true. 

What are the reasons to change this?

sbacelic:
Ah, then _TraceableProducer::class_ service needs to be made public to make it work with Symfony ^4 as:

> Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException : The "Enqueue\Client\TraceableProducer" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.

Shall I submit new PR with this change?

sbacelic:
Did in #422

PR #427, opened 2018-04-18
Allow to enable SSL in StompConnectionFactory

arjanvdbos:
Allow to enable SSL, depend the uri scheme on that. Fully backwards compatible.


makasim:
@arjanvdbos Thank you

PR #432, opened 2018-04-26
Kafka symfony transport

dheineman:
Adds the TransportFactory and Driver required to use kafka in the Symfony bundle


pkg/rdkafka/Client/RdKafkaDriver.php
```diff
@@ -0,0 +1,154 @@
...
+        return $this->context->createQueue($transportName);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setupBroker(LoggerInterface $logger = null)
+    {
+        $logger = $logger ?: new NullLogger();
+        $logger->debug('[RdKafkaDriver] setup broker');
```
dheineman: not sure if this is going to be needed.

pkg/rdkafka/Symfony/RdKafkaTransportFactory.php
```diff
@@ -0,0 +1,139 @@
...
+                })
+            ->end()
+            ->fixXmlConfig('topic')
+            ->children()
+                ->scalarNode('dsn')
+                    ->info('The kafka DSN. Other parameters are ignored if set')
+                ->end()
+                ->arrayNode('global')
+                    ->children()
+                        ->scalarNode('metadata.broker.list')->end()
```
dheineman: Symfony does not allow dots in configuration nodes but the [librdkafka configuration keys](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md) do contain them, need to figure out how to maybe normalize them?

pkg/rdkafka/Symfony/RdKafkaTransportFactory.php
```diff
@@ -0,0 +1,139 @@
...
+    {
+        $this->name = $name;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addConfiguration(ArrayNodeDefinition $builder)
+    {
+        $builder
```
dheineman: The entire configuration needs more work/checking

pkg/rdkafka/Symfony/RdKafkaTransportFactory.php
makasim: you can define a variable node (which allows everything to be put in it).

pkg/rdkafka/Client/RdKafkaDriver.php
makasim: It will, keep it.

makasim:
It looks good to me, would be happy to merge it.

pkg/rdkafka/Symfony/RdKafkaTransportFactory.php
dheineman: As far as i can tell even a variable node requires a name and will [throw an exception](https://github.com/symfony/config/blob/v4.0.8/Definition/BaseNode.php#L42) when it contains any dots.

I can just remove the `metadata.broker.list` scalarNode and just leave the `global` ArrayNode or . Any suggestions?

dergriewatz: By default you can configure the broker over the config in this way:
```
enqueue:
    transport:
        default: 'kafka://localhost:9092'
```
So I think it is a good solution to provide a alternative config way like this:
```
enqueue:
    transport:
        kafka:
            metadata_broker_list: localhost:9092
            ...
```

The easier way is to use `->scalarNode('metadata_broker_list')->end()`

dheineman: @dergriewatz As the configuration is passed to rdkafka/librdkafka when creating a new Producer or Consumer and the global configuration properties are pretty clearly defined as having dots (check my first link) so i don't think it will work.

dheineman: I went with the variable node for now as parsing or normalizing the configuration nodes would require maintaining a map of all rdkafka configuration properties (as it consists of keys with both dots and underscores). And i am not particularly in favour of this.

makasim:
@dheineman Thank you very much

makasim:
Tagged 0.8.27 with this feature included.

PR #437, opened 2018-05-08
[gps] Add support for google/cloud-pubsub ^1.0

kfb-ts:


makasim:
Have you tested it? 


Are there breaking changes between 0.6 and 1.0

kfb-ts:
Oh that was fast, cool!
I did test it, also google doesn't list any breaking changes in the related changelogs besides 0.10.0, which is about Gapic usage, which gps doesn't do directly. (https://github.com/GoogleCloudPlatform/google-cloud-php/releases/tag/v0.46.0)

PR #442, opened 2018-05-10
[bundle] Fix panel rendering when message body is an object

thePanz:


pkg/enqueue-bundle/Resources/views/Profiler/panel.html.twig
```diff
@@ -45,20 +45,21 @@
...
+                        {{ body|length > 40 ? body|slice(0, 40) ~ '...' : body  }}
                     </span>
+                    {% if body|length > 40 %}
+                        <a class="btn btn-link text-small sf-toggle"
+                           data-toggle-selector="#message-body-{{ loop.index }}"
+                           data-toggle-alt-content="Hide body"
+                        >Show body</a>
+                        <div id="message-body-{{ loop.index }}"
+                             class="context sf-toggle-content sf-toggle-hidden">
+                            <pre>{{ body }}</pre>
```
makasim: why dont you use `profiler_dump` here?

thePanz: The 'profiler_dump' expects a Data object and fails when the $body is an object.
The change here ensures the displayed value is a json string

thePanz: @makasim any updates on this?

PR #446, opened 2018-05-16
Allow newer version of bunny

enumag:


enumag:
@makasim According to the author of bunny there were no API changes.

See https://github.com/jakubkulhan/bunny/issues/66.

The failed test doesn't seem related to my changes.

makasim:
Thanks.

PR #447, opened 2018-05-19
Sending events to different queues

kocherovf:
Hi.

Thank you for your great work with this library. It's really very useful.

I am working on some project that needs `AsyncEventDispatcher` and your library is meet all needs except of the cases when I need to send different events to the different event queues. For now, I see that only one queue may be defined in `AsyncListener` so I added some logic for storing queue name in an event and using it in the listener.

Please take a look.


makasim:
@kocherovf thank you for PR. I am not sure whether we should force people to use the custom event object. The idea is to keep it transparent for the end user.

I am more in favor of making queue configurable on container tag level. https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/async-event-dispatcher/DependencyInjection/AsyncEventsPass.php#L48

kocherovf:
Unfortunately, I am not very experienced in Symphony, most like I use some custom frameworks (don't ask me why :smile:  ) or some thin frameworks like Slim with some additional components like `enqueue/async-event-dispatcher`.

So, my idea was to make extending as simple as possible

For example, you have some default queue defined like:

```
$eventQueue = $context->createQueue('default_queue_name');
$asyncLisenter = new AsyncListener($context, $registry, $eventQueue);
```

So, all events will be sent there, but if you want to use some specific queue for some particular event you can just implement `AsyncEvent` interface with this event

```
class SomeEvent extent Event imlements AsyncEvent
{
    public function getQueueName()
    {
        return 'specific_queue_name';
    }
}
```
I also choose this way because it doesn't need many changes in the code and it doesn't break anything.

makasim:
We decided not to merge this as is, for the reasons described above.

Issue #406, opened 2018-03-15
Add documentation the processor services need to be public


daften:
By default, the symfony4 services with autoconfigure aren't public. I've noticed however that processors need to be declared publicly to work.

If the service isn't public, this is the error message I receive:
```
[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]
  The "App\Service\Queue\Processor\RunCommandProcessor" service or alias has been removed or inlined when the container was compiled. You should either make it public, or s
  top using the container directly and use dependency injection instead.
```

makasim:
@daften That's indeed would be great to fix. Could you work on a PR?

Issue #405, opened 2018-03-14
Use the auto-tagging feature for e.g. processors


daften:
Similar to how twig does it: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/TwigBundle/DependencyInjection/TwigExtension.php#L154

This line in the load function of EnqueueExtension.php should do it, but this is just one example:
```
$container->registerForAutoconfiguration(\PsrProcessor::class)->addTag('enqueue.client.processor');
```

It'd seem like a convenience for symfony4, with autowiring and autoconfigure.

makasim:
@daften That's indeed would be great to fix. Could you work on a PR?

mnavarrocarter:
What is the status of this? I use Symony's autowiring/autoconfigure feature a lot, and for the processors this cloud be a great improvement. I can definetely work on a PR for this.

makasim:
A PR is very welcome

mnavarrocarter:
Cool, I'll work in one. Where is the bundle repo though? Because the one on Github says "Read Only".

mnavarrocarter:
Done! https://github.com/php-enqueue/enqueue-bundle/pull/3

Issue #410, opened 2018-03-22
[Bundle] Enqueue\Symfony\Client\ContainerAwareProcessorRegistry expects processors to be public


dkarlovi:
With Symfony 4, all services are private by default. But `Enqueue\Symfony\Client\ContainerAwareProcessorRegistry` keeps them as a set of names which it tries to `get()` from the container.

A better approach might be to inject tagged services or alternatively use a [service locator](https://symfony.com/doc/current/service_container/service_locators.html).

makasim:
A PR that fixes it would be great

dkarlovi:
We should probably fix this first, then #409 since that one will make services private by default.

PR #452, opened 2018-05-31
Add autoconfigure for services extending PsrProcess interface

mnavarrocarter:
This will allow to Symfony users that have autoconfigure/autowiring enabled to avoid defining
explicitly the processor service and its tag. Now, the symfony container does it automatically.

### DIFF
--- a/pkg/enqueue-bundle/Tests/Unit/DependencyInjection/EnqueueExtensionTest.php
+++ b/pkg/enqueue-bundle/Tests/Unit/DependencyInjection/EnqueueExtensionTest.php
@@ -6,8 +6,10 @@ use Enqueue\Bundle\DependencyInjection\Configuration;
 use Enqueue\Bundle\DependencyInjection\EnqueueExtension;
 use Enqueue\Bundle\Tests\Unit\Mocks\FooTransportFactory;
 use Enqueue\Bundle\Tests\Unit\Mocks\TransportFactoryWithoutDriverFactory;
+use Enqueue\Client\CommandSubscriberInterface;
 use Enqueue\Client\Producer;
 use Enqueue\Client\ProducerInterface;
+use Enqueue\Client\TopicSubscriberInterface;
 use Enqueue\Client\TraceableProducer;
 use Enqueue\Consumption\QueueConsumer;
 use Enqueue\JobQueue\JobRunner;
@@ -631,6 +633,31 @@ class EnqueueExtensionTest extends TestCase
                 'need_package' => true,
             ],
         ]], $container);
+    }
+
+    public function testShouldLoadProcessAutoconfigureChildDefinition()
+    {
+        $container = $this->getContainerBuilder(true);
+        $extension = new EnqueueExtension();
+
+        $extension->load([[
+            'client' => [],
+            'transport' => [],
+        ]], $container);
+
+        if (method_exists($container, 'registerForAutoconfiguration')) {
+            $autoconfigured = $container->getAutoconfiguredInstanceof();
+
+            self::assertArrayHasKey(CommandSubscriberInterface::class, $autoconfigured);
+            self::assertTrue($autoconfigured[CommandSubscriberInterface::class]->hasTag('enqueue.client.processor'));
+            self::assertTrue($autoconfigured[CommandSubscriberInterface::class]->isPublic());
+
+            self::assertArrayHasKey(TopicSubscriberInterface::class, $autoconfigured);
+            self::assertTrue($autoconfigured[TopicSubscriberInterface::class]->hasTag('enqueue.client.processor'));
+            self::assertTrue($autoconfigured[TopicSubscriberInterface::class]->isPublic());
+        } else {
+            $this->assertTrue(true);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


makasim:
This would work for processors that implement TopicSubscriber or CommandSubscriber interfaces. It would fail If none of them is implemented cuz in that case a developer should add extra parameters to the tag.

Could we do autotag for TopicSubscriber or CommandSubscribe interfaces only?

mnavarrocarter:
Okay I get it. Question is, can anyone use TopicSubscriber or CommandSubscriber without implementing PsrProcessor? I did the test and I can. Maybe what we can do is  what you say plus make CommandSubscriber and TopicSubscriber extend from the PsrProcessor interface. Just a suggestion.

mnavarrocarter:
@makasim Have you tought about what I said? If you don't think is a good idea let me know and I can just do what you say. :)

makasim:
> plus make CommandSubscriber and TopicSubscriber extend from the PsrProcessor interface

@mnavarrocarter I dont have any particular reasons against it though I tend to keep it as is. Can we do it later ?

mnavarrocarter:
@makasim Sure we can. I'll work on the autoconfig of the two interfaces you mentioned. Should be ready in a couple of hours.

mnavarrocarter:
@makasim I think we are done here. I also kept in mind #410 and made the services public. I know is not best practice, but until you guys come with a better solution, that's what it is. :)

mnavarrocarter:
Also, just a side note. I tend to squash all my commits before merging, but your merging strategy is "merge commit". If you use "fast-forward" merge you'll get a better history for feature branches that are squashed in a single commit, and you would get rid of those merge commits in the middle. :)

mnavarrocarter:
@makasim Here we go again!

makasim:
The last thing. There are code style issues, Could you please fix them https://travis-ci.org/php-enqueue/enqueue-dev/jobs/389255972?

mnavarrocarter:
I think that we are done here! That PR was tougher than I thought!

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/406, https://github.com/php-enqueue/enqueue-dev/issues/410

daften:
Thanks @mnavarrocarter  for this. I tried the PR, but didn't have time (and aren't skilled enough with symfony) to finish it in time. @makasim too. You guys make open source rock!

pkg/enqueue-bundle/Tests/Unit/DependencyInjection/EnqueueExtensionTest.php
```diff
@@ -633,6 +635,31 @@ public function testShouldThrowIfPackageShouldBeInstalledToUseTransport()
...
+    {
+        $container = $this->getContainerBuilder(true);
+        $extension = new EnqueueExtension();
+
+        $extension->load([[
+            'client' => [],
+            'transport' => [],
+        ]], $container);
+
+        if (method_exists($container, 'registerForAutoconfiguration')) {
```
makasim: The test has to be split into two. One test the case where the method is not available and whet it presents

makasim:
@mnavarrocarter could you please fix the test? I'd be happy to merge this

mnavarrocarter:
@makasim Sure I can. Sorry I've been busy at work. You'll have it fixed later today. :)

dkarlovi:
Can we merge here?

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/405

intergalactisch:
Hi guys, awesome bundle but due to the autoconfiguration I'm bumping into an issue where the Processor is cached/defined twice and therefore consumes a message twice. 

My services.yaml has autoconfigure on true, and has defined the processor as:
import_count_processor:
        class: 'App\Service\ImportCountProcessor'
        public: true
        tags:
            - { name: 'enqueue.client.processor' }
        calls:
            - [setContainer, ['@service_container']]

With this, the processor looks like this:
`<?php
namespace App\Service;

use Interop\Queue\PsrContext;
use Interop\Queue\PsrMessage;
use Interop\Queue\PsrProcessor;
use Enqueue\Client\TopicSubscriberInterface;

use Symfony\Component\DependencyInjection\ContainerAwareInterface;
use Symfony\Component\DependencyInjection\ContainerAwareTrait;
use Doctrine\ORM\EntityManagerInterface;
use Psr\Container\ContainerInterface;

class ImportCountProcessor implements TopicSubscriberInterface, ContainerAwareInterface
{

    use ContainerAwareTrait;

    protected $em;
    protected $container;

    /**
     * Importer constructor.
     *
     * @param EntityManagerInterface $em
     */
    public function __construct(EntityManagerInterface $em, ContainerInterface $container)
    {
        $this->em = $em;
        $this->container = $container;
    }

    public function process(PsrMessage $message, PsrContext $session)
    {
        echo "Import Count" . PHP_EOL;

        return self::ACK;
    }

    public static function getSubscribedTopics()
    {
        return ['importCountTopic'];
    }
}
`

Any ideas on how I can fix the duplication of the processor? Do I just need to remove the import_count_processor definition?

makasim:
set `autoconfigure: false` and keep your tag or remove your tag and let Symfony tag it.

intergalactisch:
Okido thanks for the quick reply, removing the tag from services.yaml fixed my issue

PR #453, opened 2018-05-31
PHP_URL_SCHEME doesn't support underscores

coudenysj:
https://github.com/php-enqueue/dbal/pull/2

### DIFF
--- a/pkg/dbal/DbalConnectionFactory.php
+++ b/pkg/dbal/DbalConnectionFactory.php
@@ -98,10 +98,10 @@ class DbalConnectionFactory implements PsrConnectionFactory
             throw new \LogicException(sprintf('Failed to parse DSN "%s"', $dsn));
         }
 
-        $schema = parse_url($dsn, PHP_URL_SCHEME);
-        if (empty($schema)) {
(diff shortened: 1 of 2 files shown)


coudenysj:
Is it ok if I replace it with a regex check?

makasim:
> Is it ok if I replace it with a regex check?

what exactly?

makasim:
@coudenysj ping

coudenysj:
The following line (parse_url) doesn't support underscores:

https://github.com/php-enqueue/enqueue-dev/blob/9cea3918ee88d0156d5bb9c60cef490289036498/pkg/dbal/DbalConnectionFactory.php#L101

So only a subset of this list is actually supported:

https://github.com/php-enqueue/enqueue-dev/blob/9cea3918ee88d0156d5bb9c60cef490289036498/pkg/dbal/DbalConnectionFactory.php#L106-L121

coudenysj:
Here you go.

coudenysj:
Could you re-run the checks? I didn't change anything to the MongoDB testCouldBeConstructedWithCustomConfiguration.

makasim:
@coudenysj Mongodb failed test is not related to your changes. I am aware of it.

PR #467, opened 2018-06-24
[consumption] add process niceness extension

### LINKED ISSUE #449: [consumption] process niceness extension
With PHP it is doable with just one line:
```
proc_nice($priority);
```
Where priority is integer from -20 to 19, but only values from 0 onwards is allowed without root permissions. Default is 0.

http://php.net/manual/en/function.proc-nice.php

Would be very usefull for environments with limited CPU power and where (in terms of CPU time) some consumers must overcome other not so important processes.

ramunasd:
fix #449

### DIFF
--- a/pkg/enqueue/Consumption/Extension/NicenessExtension.php
+++ b/pkg/enqueue/Consumption/Extension/NicenessExtension.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Enqueue\Consumption\Extension;
+
+use Enqueue\Consumption\Context;
+use Enqueue\Consumption\EmptyExtensionTrait;
+use Enqueue\Consumption\ExtensionInterface;
+
+class NicenessExtension implements ExtensionInterface
+{
+    use EmptyExtensionTrait;
+
+    /**
+     * @var int
+     */
+    protected $niceness = 0;
+
+    /**
+     * @param int $niceness
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function __construct($niceness)
+    {
+        if (false === is_int($niceness)) {
+            throw new \InvalidArgumentException(sprintf(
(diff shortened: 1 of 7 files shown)


pkg/enqueue/Consumption/Extension/NicenessExtension.php
```diff
@@ -0,0 +1,36 @@
...
+    public function __construct($niceness)
+    {
+        if (false === is_int($niceness)) {
+            throw new \InvalidArgumentException(sprintf(
+                'Expected niceness value is int but got: "%s"',
+                is_object($niceness) ? get_class($niceness) : gettype($niceness)
+            ));
+        }
+
+        if (0 !== $niceness) {
```
makasim: this has to be done `onStart`

pkg/enqueue/Symfony/Consumption/LimitsExtensionsCommandTrait.php
```diff
@@ -58,6 +60,11 @@ protected function getLimitsExtensions(InputInterface $input, OutputInterface $o
             $extensions[] = new LimitConsumerMemoryExtension($memoryLimit);
         }
 
+        $niceness = (int) $input->getOption('niceness');
```
makasim: we should not force int but check if the option is number and if not throw an exception with the message that explains it shuld be int.

ramunasd: The check already is in constructor of extension. Do we really need additional one?

pkg/enqueue/Consumption/Extension/NicenessExtension.php
ramunasd: fixed

PR #479, opened 2018-07-26
Fix AMQPContext::unsubscribe

adrienbrault:


adrienbrault:
I ran phpcsfixer locally but it skips the file.

makasim:
Could you try to do 

```
./bin/pre-commit -i
```

as described here https://github.com/php-enqueue/enqueue-dev/blob/master/docs/contribution.md#setup-environment

Most likely you would need to amend your commit to be able to trigger phpcsfixer hook.

PR #497, opened 2018-08-13
[redis] Authentication support added 

makasim:


Toflar:
Hey, not sure I'm on the right track here but I just wanted to use redis with a password and had the same issue as the others. Great we can now configure it directly! I was about to give this a try but it doesn't work with Predis because of the `initCommands`. You named the parameter `pass` which will cause Predis to execute `SELECT` before `AUTH` (see https://github.com/nrk/predis/blob/v1.1/src/Connection/Factory.php#L172) and then it doesn't work due to `SELECT failed: NOAUTH Authentication required` 😄

makasim:
@Toflar could you please open a new issue, or better a PR that fixes it.

Toflar:
Done.

PR #508, opened 2018-08-16
Allow either subscribe or assign in RdKafkaConsumer

Engerim:
With `assign` it is not possible to use the rebalancing from Kafka. So if no offset is set we use `subscribe` and rebalancing is possible and otherwise if a offset is set we are using assign.


mente:
@makasim when do you plan to release this bug to stable?

makasim:
@mente not sure it happens in near future. I think a backport for 0.8 is a better option.

makasim:
I was almost ready to release a stable version (3-5 days left). Now I am involved in a project development. Not sure when it happens.

PR #509, opened 2018-08-20
Fixed password auth for predis

Toflar:
As requested in https://github.com/php-enqueue/enqueue-dev/pull/497#issuecomment-414077792.
Unfortunately I'm very new to enqueue so I need some guidance here maybe on where to add tests for that 😄


makasim:
@Toflar thanks!

Toflar:
Cool thx, any release plans? I mean, I can require dev-master, no problem but the subtree split is missing 😄

makasim:
@Toflar done

PR #515, opened 2018-08-24
allow tls as scheme in DSN string for predis library

jeff1985:
Predis allows connecting to TLS encrypted endpoints by using scheme=tls which is not allowed by current enqueue library. This is needed to connect to major cloud provider's redis services. I added the posibility to specify tls:// endpoints in DSN string for each configuration.

See predis docs for more info on TLS: https://github.com/nrk/predis

My DSN string looks as following:

    ENQUEUE_DSN=tls://xxx.redis.cache.windows.net:6380?password=yyyy&ssl[verify_peer]=1&persistent=1&vendor=predis

### DIFF
--- a/pkg/redis/RedisConnectionFactory.php
+++ b/pkg/redis/RedisConnectionFactory.php
@@ -50,11 +50,7 @@ class RedisConnectionFactory implements PsrConnectionFactory
         }
 
         $this->config = array_replace($this->defaultConfig(), $config);
-        if (isset($this->config['vendor'])) {
-            $vendor = $this->config['vendor'];
-        } else {
-            $vendor = "";
-        }
+        $vendor = $this->config['vendor'];
 
         $supportedVendors = ['predis', 'phpredis', 'custom'];
         if (false == in_array($vendor, $supportedVendors, true)) {
--- a/pkg/redis/RedisConnectionFactory.php
+++ b/pkg/redis/RedisConnectionFactory.php
@@ -118,7 +118,7 @@ class RedisConnectionFactory implements PsrConnectionFactory
     {
         $unsupportedError = 'The given DSN "%s" is not supported. Must start with "redis:".';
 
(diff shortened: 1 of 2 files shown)


makasim:
Tls is to broad, I’d suggest changing it to  scheme `redis+tls`

jeff1985:
Hi @makasim ,

just tested: redis+tls throws exception in Predis library:
    
    Unknown connection scheme: 'redis+tls'

another possibility would be [rediss](http://www.iana.org/assignments/uri-schemes/prov/rediss) which is alias of `tls`. I just tested and it just works fine.

Agree?

makasim:
Agree

jeff1985:
OK @makasim, I changed it to `rediss`.

Please accept the MR!

makasim:
Could you please add some tests to RedisConnectionFactoryConfigTest?

jeff1985:
Hi @makasim,

I added tests to RedisConnectionFactoryConfigTest as you requested!

pkg/redis/RedisConnectionFactory.php
```diff
@@ -50,13 +50,18 @@ public function __construct($config = 'redis:')
         }
 
         $this->config = array_replace($this->defaultConfig(), $config);
+        if (isset($this->config['vendor'])) {
```
makasim: Could you explain what this `if ... else` for? The `defaultConfig` method returns `vendor` option, so it has to be present all the time.

jeff1985: good point! I added that line because of the check in parseDsn() method (there the parameter is not present for some of the tests as it is loaded only from the dsn string). 
in __construct the check is indeed not required. I'll remove this line.

jeff1985:
Hi @makasim 

were you able to finish your review?

jeff1985:
Thanks for that!

PR #519, opened 2018-08-27
Sqs decorator friendly

kbsali:


makasim:
@kbsali what is the problem you are trying to solve?

kbsali:
I need to use a custom `SqsContext` (decorating the Enqueue one) so I can overwrite the `getQueueUrl()`; but `SqsProducer` has a hard dependency with `Enqueue\Sqs`SqsContext`... my `getQueueUrl()` is therefore never called.
(This a Symfony integration)

makasim:
@kbsali Could you tell what you've changed in getQueueUrl()?

makasim:
@kbsali could you post the changes you are doing so I can think of a better approach

makasim:
@kbsali If you could overwrite only a factory, would it help you to solve your issue? A new bundle version allows to configure `connection_factory` class.

```yaml
enqueue:
  transport:
    dsn: '%env(ENQUEUE_DSN)%'
    connection_factory_class: 'Your/Custom/ConnectionFactory'
```

kbsali:
sorry, i'll close this PR as i'm not sure how to make it "the way you'd like" :)
We have finally managed to migrate all our queues from Redis to SQS,
but the problem remains : right now it is not possible, as far as I could see, to manage several queues with different brokers!

makasim:
@kbsali could you post the code you currently use, a workaround.

PR #533, opened 2018-09-11
Fix resolving DSN from dynamic environment variables

mdeboer:
There was a hack in the `DefaultTransportFactory ` class named `resolveDSN` which allowed for basic environment variable resolving but not for dynamic resolving.

E.g. this would work:
```
ENQUEUE_DSN=file:///tmp

enqueue:
  transport:
    default: '%env(ENQUEUE_DSN)%'
  client: ~
```

But this wouldn't:
```
ENQUEUE_DSN=file:///%kernel.cache_dir%/enqueue

enqueue:
  transport:
    default: '%env(resolve:ENQUEUE_DSN)%'
  client: ~
```

With this fix this now works like it should.

I fixed it by moving the logic of creating connection factories, contexts and drivers to a compiler pass which allows for dynamic resolving of environment placeholders.

Because this was normally done in the extension itself, some tests had to be altered to run the compiler pass after registering the extension.

### DIFF
--- a/pkg/enqueue-bundle/DependencyInjection/Compiler/BuildTransportFactoriesPass.php
+++ b/pkg/enqueue-bundle/DependencyInjection/Compiler/BuildTransportFactoriesPass.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Enqueue\Bundle\DependencyInjection\Compiler;
+
+use Enqueue\Bundle\DependencyInjection\EnqueueExtension;
+use Enqueue\Symfony\DriverFactoryInterface;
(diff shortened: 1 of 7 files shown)


makasim:
The problem is properly fixed in 0.9 see https://github.com/php-enqueue/enqueue-dev/pull/510 (current master). 

The idea of resolving env vars at compile time is not right IMO.

mdeboer:
@makasim as I need the functionality, when will 0.9 be released?

makasim:
The dev branch is pretty much stable, though some BC breaks have been done and expected to be done. I still have a lot on my plat, do not expect a release soon.

PR #534, opened 2018-09-13
[client] Introduce routes. Foundation for multi transport support. 

makasim:


makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/538

Issue #312, opened 2018-01-04
Add ability to choose transport\context to be used in consume command 


makasim:

makasim:
block https://github.com/php-enqueue/enqueue-elastica-bundle/pull/8

PR #548, opened 2018-10-02
[consumption] Add ability to consume from multiple transports.

makasim:


makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/312

Issue #549, opened 2018-10-03
Stomp heartbeat


versh23:
Hi

Sometimes activeMQ close connection with my consumer. i would to check a connection status but how? 
Connection instance **private** https://github.com/php-enqueue/stomp/blob/master/StompContext.php#L189 and there are now heartbeat config 
https://github.com/php-enqueue/stomp/blob/master/StompConnectionFactory.php#L122

so, is it possible to send isAlive package or check connection ?

makasim:
That could be made public.

versh23:
@makasim so, what can i do ? 
stomp repo is [READ-ONLY], i cant make PR ?

makasim:
You are supposed to do a PR on this repository, It will go to read only repo once is merged

makasim:
https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/stomp/StompContext.php

PR #552, opened 2018-10-03
[STOMP] make getStomp public

versh23:
related #549


makasim:
@versh23 Could you please add a test?

versh23:
@makasim done
its ok ?

Issue #323, opened 2018-01-11
[consumption] Some improvements


makasim:
* It should be possible to set AMQP's consumer tag before starting a consumption.
* It should be possible to check the broker is up and running onStart, or onEarlyStart.

PR #554, opened 2018-10-04
[consumption] Rework QueueConsumer extension points.

makasim:
* It should be possible to set AMQP's consumer tag before starting a consumption.
* It should be possible to check the broker is up and running onStart, or onEarlyStart. 

fixes https://github.com/php-enqueue/enqueue-dev/issues/323


makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/323

PR #561, opened 2018-10-16
Fixing Kafka Default configuration merge

adumas37:
Fixing the constructor to merge recursively default config and user config

PR #562, opened 2018-10-16
Fixing kafka default configuration

adumas37:
see https://github.com/php-enqueue/enqueue-dev/pull/561


adumas37:
Thank you @makasim !

PR #563, opened 2018-10-17
Dbal Subscription Consumer feature

rosamarsky:


pkg/dbal/DbalContext.php
```diff
@@ -126,7 +125,31 @@ public function close(): void
...
-        throw SubscriptionConsumerNotSupportedException::providerDoestNotSupportIt();
+        return new DbalSubscriptionConsumer($this);
+    }
+
+    /**
+     * @internal It must be used here and in the consumer only
+     */
+    public function convertMessage(array $dbalMessage): DbalMessage
+    {
+        $dbalMessageObj = new DbalMessage(
```
makasim: $this->createMessage

PR #565, opened 2018-10-18
MongoDB Subscription Consumer feature

rosamarsky:


pkg/mongodb/MongodbMessage.php
```diff
@@ -65,6 +65,22 @@ public function __construct(string $body = '', array $properties = [], array $he
         $this->redelivered = false;
     }
 
+    public static function fromArrayDbResult(array $arrayResult): self
```
makasim: move the method to the context, make it public with `@internal` docblock, look at AMQP cotext

makasim:
```suggestion
    public static function fromArrayDbResult(array $arrayResult): self
```

rosamarsky: ok, done!

pkg/mongodb/MongodbContext.php
```diff
@@ -107,7 +106,26 @@ public function close(): void
...
-        throw SubscriptionConsumerNotSupportedException::providerDoestNotSupportIt();
+        return new MongodbSubscriptionConsumer($this);
+    }
+
+    /**
+     * @internal It must be used here and in the consumer only
+     */
+    public function convertMessage(array $mongodbMessage): MongodbMessage
+    {
+        $mongodbMessageObj = new MongodbMessage(
```
makasim: $this->createMessage()

makasim:
could you please also update the transport doc like this one https://github.com/php-enqueue/enqueue-dev/blob/master/docs/transport/amqp.md#subscription-consumer

PR #581, opened 2018-10-23
[dbal] Introduce redelivery support based on visibility approach. 

rosamarsky:
* [x] Upd docs

### DIFF
--- a/pkg/dbal/DbalConsumer.php
+++ b/pkg/dbal/DbalConsumer.php
@@ -87,37 +87,19 @@ class DbalConsumer implements Consumer
 
     public function receiveNoWait(): ?Message
     {
+        $deliveryId = (string) Uuid::uuid1();
+        $redeliveryDelay = $this->getRedeliveryDelay() / 1000; // milliseconds to seconds
+
         $this->redeliverMessages();
 
-        $this->getConnection()->beginTransaction();
-        try {
-            $deliveryId = (string) Uuid::uuid1();
-
-            // get top message from the queue
-            $message = $this->fetchMessage([$this->queue->getQueueName()]);
-
-            if (null == $message) {
-                $this->getConnection()->commit();
-
-                return null;
+        // get top message from the queue
+        if ($message = $this->fetchMessage([$this->queue->getQueueName()], $deliveryId, $redeliveryDelay)) {
+            if ($message['redelivered'] || empty($message['time_to_live']) || $message['time_to_live'] > time()) {
+                return $this->getContext()->convertMessage($message);
             }
-
-            $this->markMessageAsDeliveredToConsumer($message, $deliveryId);
-
-            $dbalMessage = $this->getMessageByDeliveryId($deliveryId);
-
-            $this->getConnection()->commit();
-
-            if ($dbalMessage['redelivered'] || empty($dbalMessage['time_to_live']) || $dbalMessage['time_to_live'] > time()) {
-                return $this->getContext()->convertMessage($dbalMessage);
-            }
-
-            return null;
-        } catch (\Exception $e) {
-            $this->dbal->rollBack();
-
-            throw $e;
         }
+
+        return null;
     }
 
     /**
--- a/pkg/dbal/DbalConsumer.php
+++ b/pkg/dbal/DbalConsumer.php
@@ -6,13 +6,18 @@ namespace Enqueue\Dbal;
 
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Types\Type;
+use function GuzzleHttp\Psr7\str;
 use Interop\Queue\Consumer;
 use Interop\Queue\Exception\InvalidMessageException;
+use Interop\Queue\Impl\ConsumerPollingTrait;
 use Interop\Queue\Message;
 use Interop\Queue\Queue;
+use Ramsey\Uuid\Uuid;
 
 class DbalConsumer implements Consumer
 {
+    use ConsumerPollingTrait;
+
     /**
      * @var DbalContext
      */
@@ -29,9 +34,11 @@ class DbalConsumer implements Consumer
     private $queue;
 
     /**
+     * Default 20 minutes in milliseconds.
+     *
      * @var int
      */
-    private $pollingInterval;
+    private $redeliveryDelay;
 
     public function __construct(DbalContext $context, DbalDestination $queue)
     {
@@ -39,23 +46,25 @@ class DbalConsumer implements Consumer
         $this->queue = $queue;
         $this->dbal = $this->context->getDbalConnection();
 
-        $this->pollingInterval = 1000;
+        $this->redeliveryDelay = 1200000;
     }
 
     /**
-     * Polling interval is in milliseconds.
+     * Get interval between retry failed messages in milliseconds.
      */
-    public function setPollingInterval(int $interval): void
+    public function getRedeliveryDelay(): int
     {
-        $this->pollingInterval = $interval;
+        return $this->redeliveryDelay;
     }
 
     /**
-     * Get polling interval in milliseconds.
+     * Interval between retry failed messages in seconds.
      */
-    public function getPollingInterval(): int
+    public function setRedeliveryDelay(int $redeliveryDelay): self
     {
-        return $this->pollingInterval;
+        $this->redeliveryDelay = $redeliveryDelay;
+
+        return $this;
     }
 
     /**
@@ -66,33 +75,59 @@ class DbalConsumer implements Consumer
         return $this->queue;
     }
 
-    public function receive(int $timeout = 0): ?Message
-    {
-        $timeout /= 1000;
-        $startAt = microtime(true);
-
-        while (true) {
-            $message = $this->receiveMessage();
-
-            if ($message) {
-                return $message;
-            }
-
-            if ($timeout && (microtime(true) - $startAt) >= $timeout) {
-                return null;
-            }
-
-            usleep($this->pollingInterval * 1000);
-
-            if ($timeout && (microtime(true) - $startAt) >= $timeout) {
-                return null;
-            }
-        }
-    }
-
     public function receiveNoWait(): ?Message
     {
-        return $this->receiveMessage();
+        $this->redeliverMessages();
+
+        $this->dbal->beginTransaction();
+        try {
+            $now = (int) time();
+            $redeliveryDelay = $this->getRedeliveryDelay() / 1000; // milliseconds to seconds
+            $deliveryId = (string) Uuid::uuid1();
+
+            // get top message from the queue
+            $message = $this->fetchMessage($now);
+
+            if (null == $message) {
+                $this->dbal->commit();
+
+                return null;
+            }
+
+            // mark message as delivered to consumer
+            $this->dbal->createQueryBuilder()
+                ->update($this->context->getTableName())
+                ->set('delivery_id', ':deliveryId')
+                ->set('redeliver_after', ':redeliverAfter')
+                ->andWhere('id = :id')
+                ->setParameter('id', $message['id'], Type::GUID)
+                ->setParameter('deliveryId', $deliveryId, Type::STRING)
+                ->setParameter('redeliverAfter', $now + $redeliveryDelay, Type::BIGINT)
+                ->execute()
+            ;
+
+            $dbalMessage = $this->dbal->createQueryBuilder()
+                ->select('*')
+                ->from($this->context->getTableName())
+                ->andWhere('delivery_id = :deliveryId')
+                ->setParameter('deliveryId', $deliveryId, Type::STRING)
+                ->setMaxResults(1)
+                ->execute()
+                ->fetch()
+            ;
+
+            $this->dbal->commit();
+
+            if ($message->isRedelivered() || empty($dbalMessage['time_to_live']) || $dbalMessage['time_to_live'] > time()) {
+                return $this->context->convertMessage($dbalMessage);
+            }
+
+            return null;
+        } catch (\Exception $e) {
+            $this->dbal->rollBack();
+
+            throw $e;
+        }
     }
 
     /**
@@ -100,7 +135,7 @@ class DbalConsumer implements Consumer
      */
     public function acknowledge(Message $message): void
     {
-        // does nothing
+        $this->deleteMessage($message->getDeliveryId());
     }
 
     /**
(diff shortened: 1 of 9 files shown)


pkg/dbal/DbalConsumer.php
```diff
@@ -33,13 +39,25 @@ class DbalConsumer implements Consumer
...
 
-    public function __construct(DbalContext $context, DbalDestination $queue)
+    /**
+     * @var int
+     */
+    private $redeliveryDelay;
+
+    public function __construct(UuidInterface $uuid, DbalContext $context, DbalDestination $queue)
     {
+        $this->uuid = $uuid;
```
makasim: consumerId

pkg/dbal/DbalConsumer.php
```diff
@@ -33,13 +39,25 @@ class DbalConsumer implements Consumer
...
         $this->context = $context;
         $this->queue = $queue;
         $this->dbal = $this->context->getDbalConnection();
 
         $this->pollingInterval = 1000;
+        $this->redeliveryDelay = 600; // 10 minutes
+    }
+
+    public function getUuid(): UuidInterface
```
makasim: getConsumerId(): string

pkg/dbal/DbalConsumer.php
```diff
@@ -33,13 +39,25 @@ class DbalConsumer implements Consumer
...
+
+    public function __construct(UuidInterface $uuid, DbalContext $context, DbalDestination $queue)
     {
+        $this->uuid = $uuid;
         $this->context = $context;
         $this->queue = $queue;
         $this->dbal = $this->context->getDbalConnection();
 
         $this->pollingInterval = 1000;
+        $this->redeliveryDelay = 600; // 10 minutes
```
makasim: should be in milliseconds.  20 minutes looks better to me.

pkg/dbal/DbalConsumer.php
```diff
@@ -58,6 +76,22 @@ public function getPollingInterval(): int
...
+    {
+        return $this->redeliveryDelay;
+    }
+
+    /**
+     * Interval between retry failed messages in seconds.
+     */
+    public function setRedeliveryDelay(int $redeliveryDelay): void
+    {
+        $this->redeliveryDelay = $redeliveryDelay;
```
makasim: should return `$this`. return type hint should be `self`

pkg/dbal/DbalConsumer.php
```diff
@@ -115,31 +149,49 @@ public function reject(Message $message, bool $requeue = false): void
 
             return;
         }
+
+        $this->deleteMessageByCustomerId($message->getConsumerId());
     }
 
     protected function receiveMessage(): ?DbalMessage
     {
+        $this->refreshRedeliveredMessages();
```
makasim: $this->redeliverMessages();

pkg/dbal/DbalConsumer.php
```diff
@@ -115,33 +97,54 @@ public function reject(Message $message, bool $requeue = false): void
 
             return;
         }
+
+        $this->deleteMessageByDeliveryId($message->getDeliveryId());
```
makasim: rename to deleteMessage. the method should accept a message instance and throws an exception if delivery id is not set.

pkg/dbal/DbalConsumer.php
```diff
@@ -115,33 +97,54 @@ public function reject(Message $message, bool $requeue = false): void
...
+
         $this->dbal->beginTransaction();
         try {
-            $now = time();
-
-            $dbalMessage = $this->fetchPrioritizedMessage($now) ?: $dbalMessage = $this->fetchMessage($now);
-            if (false == $dbalMessage) {
+            $now = (int) time();
+            $redeliveryDelay = $this->getRedeliveryDelay() / 1000; // milliseconds to seconds
+            $deliveryId = Uuid::uuid1()->toString();
```
makasim: I'd prefer `(string) Uuid::uuid1()`

pkg/dbal/DbalConsumer.php
```diff
@@ -115,33 +97,54 @@ public function reject(Message $message, bool $requeue = false): void
...
+
+            // get top message from the queue
+            $message = $this->fetchMessage($now);
```
makasim: the method could return null, if no message in the queue. but you do not check that a few lines later and assume the message is there.

pkg/dbal/DbalConsumer.php
```diff
@@ -115,33 +97,54 @@ public function reject(Message $message, bool $requeue = false): void
...
+
+            // mark message as delivered to consumer
+            $updateResult = $this->dbal->createQueryBuilder()
+                ->update($this->context->getTableName())
+                ->set('delivery_id', ':deliveryId')
+                ->set('redeliver_after', ':redeliverAfter')
+                ->andWhere('id = :id')
+                ->setParameter('id', $message['id'], Type::GUID)
+                ->setParameter('deliveryId', $deliveryId, Type::STRING)
```
makasim: make sure there is a unique index for this column.

pkg/dbal/DbalConsumer.php
```diff
@@ -115,33 +97,54 @@ public function reject(Message $message, bool $requeue = false): void
...
+                ->setMaxResults(1)
+                ->execute()
+                ->fetch()
+            ;
 
             $this->dbal->commit();
 
-            if (empty($dbalMessage['time_to_live']) || ($dbalMessage['time_to_live'] / 1000) > microtime(true)) {
+            if (empty($dbalMessage['time_to_live']) || $dbalMessage['time_to_live'] > time()) {
```
makasim: do not expire redelivered messages

pkg/dbal/DbalContext.php
```diff
@@ -212,13 +223,16 @@ public function createDataBaseTable(): void
...
+        $table->addColumn('delivery_id', Type::STRING, ['notnull' => false]);
+        $table->addColumn('redeliver_after', Type::BIGINT, ['notnull' => false]);
 
         $table->setPrimaryKey(['id']);
         $table->addIndex(['published_at']);
         $table->addIndex(['queue']);
         $table->addIndex(['priority']);
         $table->addIndex(['delayed_until']);
         $table->addIndex(['priority', 'published_at']);
+        $table->addIndex(['delivery_id']);
```
makasim: this should a unique index.

makasim: add an index for `redeliver_after`.

pkg/dbal/DbalContext.php
```diff
@@ -212,13 +223,18 @@ public function createDataBaseTable(): void
...
 
         $table->setPrimaryKey(['id']);
         $table->addIndex(['published_at']);
         $table->addIndex(['queue']);
         $table->addIndex(['priority']);
         $table->addIndex(['delayed_until']);
         $table->addIndex(['priority', 'published_at']);
+        $table->addIndex(['redeliver_after']);
+        $table->addUniqueIndex(['delivery_id']);
+        $table->addIndex(['delivery_id']);
```
makasim: Remove NOT unique index ( there is no need for two indexes)

pkg/dbal/DbalConsumer.php
```diff
@@ -115,102 +149,51 @@ public function reject(Message $message, bool $requeue = false): void
 
             return;
         }
+
+        $this->deleteMessage($message->getDeliveryId());
     }
 
-    protected function receiveMessage(): ?DbalMessage
+    private function deleteMessage(?string $deliveryId): void
```
makasim: Null is not allowed.

makasim: I would also add a check that the string is not empty.

pkg/dbal/DbalConsumer.php
```diff
@@ -115,102 +149,51 @@ public function reject(Message $message, bool $requeue = false): void
...
             ->select('*')
             ->from($this->context->getTableName())
+            ->andWhere('delivery_id IS NULL')
             ->andWhere('queue = :queue')
-            ->andWhere('priority IS NOT NULL')
-            ->andWhere('(delayed_until IS NULL OR delayed_until <= :delayedUntil)')
-            ->addOrderBy('published_at', 'asc')
+            ->andWhere('delayed_until IS NULL OR delayed_until <= :delayedUntil')
             ->addOrderBy('priority', 'desc')
+            ->addOrderBy('published_at', 'asc')
```
makasim: could you please test the behavior of sorting when a priority is null

pkg/dbal/DbalConsumer.php
```diff
@@ -115,102 +149,51 @@ public function reject(Message $message, bool $requeue = false): void
...
-            ->select('*')
-            ->from($this->context->getTableName())
-            ->andWhere('queue = :queue')
-            ->andWhere('priority IS NULL')
-            ->andWhere('(delayed_until IS NULL OR delayed_until <= :delayedUntil)')
-            ->addOrderBy('published_at', 'asc')
-            ->setMaxResults(1)
+        $this->dbal->createQueryBuilder()
+            ->update($this->context->getTableName())
+            ->set('delivery_id', ':deliveryId')
```
makasim: ->set('delivery_id', null)

pkg/dbal/DbalConsumer.php
```diff
@@ -115,102 +149,51 @@ public function reject(Message $message, bool $requeue = false): void
...
+            ->set('redelivered', ':redelivered')
```
makasim: ->set('redelivered', true)

makasim:
@rosamarsky could you update docs? use as an example https://github.com/php-enqueue/enqueue-dev/pull/585/files#diff-1ab2e57b012481304b916b4987c773b5L8

makasim:
The code shared between Consumer and SubscriptionConsumer could go to a trait. Look at how it is done in redis https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/redis/RedisConsumerHelperTrait.php

pkg/dbal/DbalConsumer.php
```diff
@@ -66,41 +85,47 @@ public function getQueue(): Queue
...
+
+            if (null == $message) {
+                $this->getConnection()->commit();
 
-            if ($timeout && (microtime(true) - $startAt) >= $timeout) {
                 return null;
             }
 
-            usleep($this->pollingInterval * 1000);
+            $this->markMessageAsDeliveredToConsumer($message, $deliveryId);
```
makasim: `markMessageDelivered`

pkg/dbal/DbalConsumerHelperTrait.php
```diff
@@ -0,0 +1,67 @@
...
+            ->fetch()
+        ;
+
+        return $result ?: null;
+    }
+
+    protected function markMessageAsDeliveredToConsumer(array $message, string $deliveryId): void
+    {
+        $now = time();
+        $redeliveryDelay = $this->getRedeliveryDelay() / 1000; // milliseconds to seconds
```
makasim: the method could not be used in the trait. pass the delay as argument

pkg/dbal/DbalConsumer.php
```diff
@@ -111,106 +136,39 @@ public function reject(Message $message, bool $requeue = false): void
...
+            ->set('redelivered', ':redelivered')
+            ->andWhere('delivery_id IS NOT NULL')
+            ->andWhere('redeliver_after < :now')
+            ->setParameter(':now', (int) time(), Type::BIGINT)
+            ->setParameter('deliveryId', null, Type::STRING)
+            ->setParameter('redelivered', true, Type::BOOLEAN)
+            ->execute()
         ;
-
-        $sql = $query->getSQL().' '.$this->dbal->getDatabasePlatform()->getWriteLockSQL();
```
makasim: we need a transaction in both consumer and subscription consumer 
both should use getWriteLockSQL

pkg/dbal/DbalConsumerHelperTrait.php
```diff
@@ -0,0 +1,67 @@
...
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Types\Type;
+
+trait DbalConsumerHelperTrait
+{
+    abstract public function getContext(): DbalContext;
+
+    abstract public function getConnection(): Connection;
+
+    protected function fetchMessage(array $queues): ?array
```
makasim: this should be the only one method, it should return an array with a message and queue it was consumed from.

pkg/dbal/DbalConsumer.php
```diff
@@ -111,106 +118,39 @@ public function reject(Message $message, bool $requeue = false): void
...
-        return $result ?: null;
+        $this->getConnection()->delete(
+            $this->getContext()->getTableName(),
+            ['delivery_id' => $deliveryId],
+            ['delivery_id' => Type::STRING]
+        );
     }
 
-    private function fetchMessage(int $now): ?array
+    private function redeliverMessages(): void
```
makasim: the method should be in the consumer trait. It must be called in subscription consumer too.

pkg/dbal/DbalConsumerHelperTrait.php
```diff
@@ -0,0 +1,78 @@
...
+    abstract public function getContext(): DbalContext;
+
+    abstract public function getConnection(): Connection;
+
+    protected function fetchMessage(array $queues, string $deliveryId, int $redeliveryDelay): ?array
+    {
+        try {
+            $now = time();
+
+            $this->getConnection()->beginTransaction();
```
makasim: this should be out of `try ... catch` block

pkg/dbal/DbalConsumerHelperTrait.php
```diff
@@ -0,0 +1,78 @@
...
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Types\Type;
+
+trait DbalConsumerHelperTrait
+{
+    abstract public function getContext(): DbalContext;
+
+    abstract public function getConnection(): Connection;
+
+    protected function fetchMessage(array $queues, string $deliveryId, int $redeliveryDelay): ?array
```
makasim: What if this method returns an instance of `DbalMessage`. You have access to context so you can use a `$context->convertMessage()` method.

pkg/dbal/DbalSubscriptionConsumer.php
```diff
@@ -43,8 +78,10 @@ public function consume(int $timeout = 0): void
             throw new \LogicException('No subscribers');
         }
 
-        $timeout = (int) ceil($timeout / 1000);
-        $endAt = time() + $timeout;
+        $now = time();
+        $timeout /= 1000;
+        $deliveryId = (string) Uuid::uuid1();
```
makasim: it should be generated per message, hence the var init should be done before fetchMessage call.

pkg/dbal/DbalSubscriptionConsumer.php
```diff
@@ -43,8 +78,10 @@ public function consume(int $timeout = 0): void
...
+        $redeliveryDelay = $this->getRedeliveryDelay() / 1000; // milliseconds to seconds
```
makasim: Already seconds

pkg/dbal/DbalSubscriptionConsumer.php
```diff
@@ -57,12 +94,8 @@ public function consume(int $timeout = 0): void
                 $currentQueueNames = $queueNames;
             }
```
makasim: should call `redeliverMessages` method. on every cycle.

pkg/dbal/DbalConsumer.php
```diff
@@ -66,41 +74,30 @@ public function getQueue(): Queue
...
+        $redeliveryDelay = $this->getRedeliveryDelay() / 1000; // milliseconds to seconds
 
-            usleep($this->pollingInterval * 1000);
+        $this->redeliverMessages();
 
-            if ($timeout && (microtime(true) - $startAt) >= $timeout) {
-                return null;
+        // get top message from the queue
+        if ($message = $this->fetchMessage([$this->queue->getQueueName()], $redeliveryDelay)) {
+            if ($message['redelivered'] || empty($message['time_to_live']) || $message['time_to_live'] > time()) {
```
makasim: the expired message has to be deleted from message, I think that expiration logic could be moved to fetchMessage method

makasim: I gave it another though, and I now I think we could cook a query that removes expired messages. just like redeliverMessage. the expiration in the code should still be preserved

pkg/dbal/DbalConsumer.php
```diff
@@ -111,106 +108,34 @@ public function reject(Message $message, bool $requeue = false): void
         InvalidMessageException::assertMessageInstanceOf($message, DbalMessage::class);
 
         if ($requeue) {
-            $this->context->createProducer()->send($this->queue, $message);
+            $this->getContext()->createProducer()->send($this->queue, $message);
```
makasim: The redelivered flag should be forced false.

pkg/dbal/DbalConsumer.php
```diff
@@ -111,106 +108,34 @@ public function reject(Message $message, bool $requeue = false): void
...
-                'delayedUntil' => Type::INTEGER,
-            ]
-        )->fetch();
-
-        return $result ?: null;
+        return $this->dbal;
     }
 
-    private function fetchMessage(int $now): ?array
+    private function deleteMessage(string $deliveryId): void
```
makasim: the method should be moved to trait.

pkg/dbal/DbalConsumer.php
```diff
@@ -108,6 +109,8 @@ public function reject(Message $message, bool $requeue = false): void
         InvalidMessageException::assertMessageInstanceOf($message, DbalMessage::class);
 
         if ($requeue) {
+            $message->setRedelivered(false);
```
makasim: clone message and than force redelivered false

makasim:
Thank you Roman for your work!

PR #585, opened 2018-10-26
Redis New Implementation

ASKozienko:
* [x] Upd docs

Supports 

* Redelivery
* Delaying
* Expiration


pkg/redis/RedisConsumer.php
```diff
@@ -21,12 +21,42 @@ class RedisConsumer implements Consumer
...
     public function __construct(RedisContext $context, RedisDestination $queue)
     {
         $this->context = $context;
         $this->queue = $queue;
     }
 
+    /**
+     * @return int
+     */
+    public function getRetryDelay(): ?int
```
makasim: redeliveryDelay

pkg/redis/RedisQueueConsumer.php
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * @var string[]
+     */
+    private $queueNames;
+
+    /**
+     * @param Redis              $redis
+     * @param RedisDestination[] $queues
+     */
+    public function __construct(Redis $redis, array $queues)
```
makasim: I'd move `array $queues` to receiveXXX methods. I think it would be better to make it as stateless as possible

pkg/redis/RedisQueueConsumer.php
```diff
@@ -0,0 +1,148 @@
...
+        $this->redis = $redis;
+        $this->queues = $queues;
+    }
+
+    /**
+     * @return int
+     */
+    public function getRetryDelay(): int
+    {
+        return $this->retryDelay;
```
makasim: could we pass it as an argument of receiveXXX methods? S

pkg/redis/RedisQueueConsumer.php
```diff
@@ -0,0 +1,148 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Enqueue\Redis;
+
+class RedisQueueConsumer
```
makasim: The name is misleading IMO. Could it be named `RedisInternalConsumer` or `RedisGenericConsumer`.

pkg/redis/RedisConsumer.php
```diff
@@ -83,15 +110,33 @@ public function reject(Message $message, bool $requeue = false): void
...
     }
 
     private function getRedis(): Redis
     {
         return $this->context->getRedis();
     }
+
+    private function initQueueConsumer(): void
+    {
+        if (null === $this->queueConsumer) {
```
makasim: if we make a RedisQueueConsumer stateless we could move it to the context and pass to real consumers as a dependency.

makasim: The init would not be needed.

pkg/redis/RedisConsumer.php
```diff
@@ -21,12 +21,42 @@ class RedisConsumer implements Consumer
...
+    /**
+     * @param int $retryDelay
+     */
+    public function setRetryDelay(int $retryDelay): void
+    {
+        $this->retryDelay = $retryDelay;
+
+        if ($this->queueConsumer) {
+            $this->queueConsumer->setRetryDelay($this->retryDelay);
+        }
```
makasim: are you sure this if condition is not buggy?

pkg/redis/RedisConsumer.php
```diff
@@ -21,12 +21,42 @@ class RedisConsumer implements Consumer
...
+    }
+
+    /**
+     * @param int $retryDelay
+     */
+    public function setRetryDelay(int $retryDelay): void
+    {
+        $this->retryDelay = $retryDelay;
+
+        if ($this->queueConsumer) {
```
makasim: I think setting retryDelay property is enough. No need to set same value on queue consumer too. (It will be passed as argument to receiveXXX methods).

pkg/redis/RedisConsumer.php
```diff
@@ -40,40 +70,37 @@ public function getQueue(): Queue
...
+        return $this->queueConsumer->receiveMessage($timeout);
     }
 
     /**
      * @return RedisMessage
      */
     public function receiveNoWait(): ?Message
     {
-        if ($result = $this->getRedis()->rpop($this->queue->getName())) {
-            return RedisMessage::jsonUnserialize($result->getMessage());
```
makasim: Could you add a serializer concept (as you've done in another transport)?

makasim: and remove serialization logic from the message.

pkg/redis/RedisMessage.php
```diff
@@ -139,6 +149,75 @@ public function getReplyTo(): ?string
...
+     */
+    public function getReservedKey(): ?string
+    {
+        return $this->reservedKey;
+    }
+
+    /**
+     * @param string $reservedKey
+     */
+    public function setReservedKey(string $reservedKey)
```
makasim: `: void`

pkg/redis/RedisQueueConsumer.php
```diff
@@ -0,0 +1,148 @@
...
+        return null;
+    }
+
+    private function processResult(RedisResult $result): ?RedisMessage
+    {
+        $message = RedisMessage::jsonUnserialize($result->getMessage());
+
+        $now = time();
+
+        if ($expiresAt = $message->getHeader('expires_at')) {
```
makasim: If we managed to processes a message before expiration time, but failed and as a result the message was redelivered, and this condition reject redelivered message as expired. I dont think this is desired behavior.

pkg/redis/RedisQueueConsumer.php
```diff
@@ -0,0 +1,148 @@
...
+
+        $reservedQueue = $result->getKey().':reserved';
+        $retryMessageAt = $now + $this->retryDelay;
+
+        $this->redis->zadd($reservedQueue, $message->getReservedKey(), $retryMessageAt);
+
+        return $message;
+    }
+
+    private function pushQueueNameBack($queueName): void
```
makasim: `string $queueName`

pkg/redis/RedisProducer.php
```diff
@@ -34,24 +45,44 @@ public function send(Destination $destination, Message $message): void
         InvalidDestinationException::assertDestinationInstanceOf($destination, RedisDestination::class);
         InvalidMessageException::assertMessageInstanceOf($message, RedisMessage::class);
 
-        $this->redis->lpush($destination->getName(), json_encode($message));
+        $message->setMessageId(UUID::generate());
```
makasim: require ramnsy uuid and use it.

PR #586, opened 2018-10-26
Allow JobStorage to reset the EntityManager

damijank:
After a Job triggers a Doctrine exception, and handles it, the EntityManager still remains closed and JobStorage can't persist changes. This fix resets the Manager after such an occurrence.


damijank:
I'm not sure if, for completeness sake, there should be a consideration about the Manager could be some other than the default.

PR #587, opened 2018-10-26
[Symfony] Try to get env parameter from container before throwing an exception

Steveb-p:
In regard to an issue raised on gitter.

image

### DIFF
--- a/pkg/enqueue/Symfony/DefaultTransportFactory.php
+++ b/pkg/enqueue/Symfony/DefaultTransportFactory.php
@@ -24,6 +24,7 @@ use Interop\Amqp\AmqpConnectionFactory;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
 use function Enqueue\dsn_to_connection_factory;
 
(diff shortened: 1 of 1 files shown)


makasim:
The issue was properly fixed in 0.9 version. Unfortunately, it is not possible to fix it in 0.8 version.  
There is another attempt to do it https://github.com/php-enqueue/enqueue-dev/pull/533

pkg/enqueue/Symfony/DefaultTransportFactory.php
```diff
@@ -161,7 +162,11 @@ private function resolveDSN(ContainerBuilder $container, $dsn)
             $matches = [];
             if (preg_match('/%env\((.*?)\)/', $dsn, $matches)) {
                 if (false === $realDsn = getenv($matches[1])) {
-                    throw new \LogicException(sprintf('The env "%s" var is not defined', $matches[1]));
+                    try {
+                        $realDsn = $container->getParameter($matches[1]);
```
makasim: shouldn't it be `$dsn` instead of `$matches[1]`

makasim: @Steveb-p ping ^^

Steveb-p: I'll take a look at it tonight :)

makasim:
@Steveb-p is there any news?

makasim:
The issue is properly fixed in 0.9 version which is soon to be released.

PR #606, opened 2018-11-01
Metrics

ASKozienko:


pkg/monitoring/WampStorage.php
```diff
@@ -0,0 +1,205 @@
...
+     *
+     * or
+     *
+     * wamp://127.0.0.1:9090?max_retries=10
+     *
+     * @param array|string|null $config
+     */
+    public function __construct($config = 'wamp:')
+    {
+        if (empty($config)) {
```
makasim: check that 3rd party library is installed. (add the lib to composer suggest section).

pkg/monitoring/StatsStorage.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Enqueue\Monitoring;
+
+interface StatsStorage
+{
+    public function pushConsumerStats(ConsumerStats $event): void;
```
makasim: add a method to track send messages stats.

pkg/monitoring/SentMessageStats.php
```diff
@@ -0,0 +1,48 @@
...
+     * @var array
+     */
+    protected $headers;
+
+    /**
+     * @var array
+     */
+    protected $properties;
+
+    public function __construct(
```
makasim: could you please add a queue or topic the message is being sent to.

makasim:
@ASKozienko some additions

* ConsumeMessageStats should have a redelivered field
* SentMessageStats should have messageId field (could be null)
* SentMessageStats should have correlationId field (could be null)

pkg/monitoring/MonitoringExtension.php
```diff
@@ -0,0 +1,315 @@
...
+                $this->getSystemLoad()
+            );
+
+            $this->safeCall(function () use ($event) {
+                $this->storage->pushConsumerStats($event);
+            }, $context->getLogger());
+        }
+    }
+
+    public function onPostConsume(PostConsume $context): void
```
makasim: You could implement only interfaces you need

makasim:
MessageStats obj should allow setting exception details too.

makasim:
* Splint Client Extension Interface, (look at Consumption Extension for reference).
* Add StorageFactory (and interface) that accept string dsn and return monitoring storage, (for now you could hard code influxdb and wamp in the factory).

pkg/monitoring/ClientMonitoringExtension.php
```diff
@@ -0,0 +1,63 @@
...
+        $timestampMs = (int) (microtime(true) * 1000);
+
+        $destination = $context->getTransportDestination() instanceof Topic
+            ? $context->getTransportDestination()->getTopicName()
+            : $context->getTransportDestination()->getQueueName()
+        ;
+
+        $stats = new SentMessageStats(
+            $timestampMs,
+            $destination,
```
makasim: add a bool filed to find out was it a topic or not.

PR #616, opened 2018-11-05
Compatibility with 0.8x

ASKozienko:
fixes https://github.com/php-enqueue/enqueue-dev/issues/598


pkg/enqueue/Client/RouterProcessor.php
```diff
@@ -21,6 +26,17 @@ public function __construct(DriverInterface $driver)
 
     public function process(InteropMessage $message, Context $context): Result
     {
+        // compatibility with 0.8x
+        if (self::COMMAND_TOPIC_08X === $message->getProperty(Config::TOPIC)) {
```
makasim: Config::TOPIC is not right property.

Issue #621, opened 2018-11-07
[Symfony] Try to check private service existence from container


Gnucki:
I was trying to migrate a bundle to enqueue 0.9 (from 0.8) in order to use symfony 4.2 messaging component instead of 4.1.

I did not changed my enqueue config:
```yaml
enqueue:
    async_events:
        enabled: false
    transport:
        default: rabbitmq_amqp
        rabbitmq_amqp:
            dsn: '%env(BROKER_DSN)%'
    client: ~
```

But got this strange exception:
> RuntimeException: Can't find Enqueue's transport named "default": Service "enqueue.transport.default.context" is not found.

Then, I tried to trace the exception. It is coming from [messenger-adapter](https://github.com/php-enqueue/messenger-adapter/blob/master/QueueInteropTransportFactory.php#L81):
```php
if (!$this->container->has($contextService = 'enqueue.transport.'.$enqueueContextName.'.context')) {
```

The problem is that context services are private so they are not accessible from container get method.

Am I missing something?
If not, I made a PR:

makasim:
context has to be added to `enqueue.locator` just like [here](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Symfony/DependencyInjection/TransportFactory.php#L190) and the locator should be used in messenger adapter (instead of the whole container). 

Services are public in the locator.

PR #622, opened 2018-11-07
Add transport context service public visibility

Gnucki:
#621


makasim:
This is not right way to address the issue, See https://github.com/php-enqueue/enqueue-dev/issues/621#issuecomment-436634190

Issue #625, opened 2018-11-07
SQS and fallback subscription consumer


deguif:
Setting receive-timeout when launching consume command does not work anymore for SQS consumer.

The `SqsContext::createSubscriptionConsumer()` method throws a `SubscriptionConsumerNotSupportedException` exception which leads to the `FallbackSubscriptionConsumer::consume()` method being called, and `receiveNoWait()` is called on the `SqsConsumer`

This make the AWS bill very high.

PR #626, opened 2018-11-08
Avoid receiveNoWait when only one subscriber

deguif:
This follows #625 I created yesterday 

Let's try to explain the issue with further details.
This occurs when using AWS SQS as queue system.

The SQS context doesn't support `SubscriptionConsumer` and throws an exception when calling its `createSubscriptionConsumer()` method. This exception is then caught and is replaced by the fallback subscription consumer.

The `FallbackSubscriptionConsumer` is a basic implementation which loops over a list of subscribers and call on each of them their `receiveNoWait()` method.
This is a cost killer for AWS SQS as this will query for new messages without polling.

Here's an example of the command I run and which leads to this dramatic cost:
`php bin/console enqueue:transport:consume --time-limit=+80seconds --receive-timeout=20000 --idle-timeout=5000 MyApp\Processor`
This command is run by a supervising system which rerun it once it stops (reached its time limit).

And here's my proposed solution, which avoids calling `receiveNoWait()` when there is only one subscriber in the fallback subscription consumer.
This is not a full fix as this won't avoid this cost killing when passing multiple subscribers to the fallback subscription consumer.

### DIFF
--- a/pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
+++ b/pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
@@ -25,13 +25,13 @@ final class FallbackSubscriptionConsumer implements SubscriptionConsumer
         $this->subscribers = [];
     }
 
-    public function consume(int $timeout = 0): void
+    public function consume(int $timeoutMs = 0): void
     {
-        if (empty($this->subscribers)) {
+        if (!$subscriberCount = \count($this->subscribers)) {
             throw new \LogicException('No subscribers');
         }
 
-        $timeout /= 1000;
+        $timeout = $timeoutMs / 1000;
         $endAt = microtime(true) + $timeout;
 
         while (true) {
@@ -41,13 +41,13 @@ final class FallbackSubscriptionConsumer implements SubscriptionConsumer
              * @var callable $processor
              */
             foreach ($this->subscribers as $queueName => list($consumer, $callback)) {
-                $message = $consumer->receiveNoWait();
+                $message = 1 === $subscriberCount ? $consumer->receive($timeoutMs) : $consumer->receiveNoWait();
 
                 if ($message) {
                     if (false === call_user_func($callback, $message, $consumer)) {
                         return;
                     }
-                } else {
+                } elseif (1 !== $subscriberCount) {
                     if ($timeout && microtime(true) >= $endAt) {
                         return;
                     }


pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
```diff
@@ -13,21 +13,16 @@
...
 
-    public function __construct()
-    {
-        $this->subscribers = [];
-    }
-
     public function consume(int $timeout = 0): void
     {
-        if (empty($this->subscribers)) {
+        if (!$subscriberCount = \count($this->subscribers)) {
```
makasim: please revert the change, the var could be initialized after this if condition.

pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
```diff
@@ -41,13 +36,13 @@ public function consume(int $timeout = 0): void
              * @var callable $processor
              */
             foreach ($this->subscribers as $queueName => list($consumer, $callback)) {
-                $message = $consumer->receiveNoWait();
+                $message = 1 === $subscriberCount ? $consumer->receive((int) ($timeout * 1000)) : $consumer->receiveNoWait();
```
makasim: [timeout should be milliseconds already](https://github.com/queue-interop/queue-interop/blob/master/src/SubscriptionConsumer.php#L9)

pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
deguif: Simple question: isn't `$this->subscribers` always an array?

makasim: always

pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
deguif: It was previously converted to seconds, but I misunderstood the previous operation done on timeout :)

pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
```diff
@@ -13,21 +13,16 @@
...
 
     /**
      * @var int
      */
     private $idleTime = 0;
 
-    public function __construct()
-    {
-        $this->subscribers = [];
-    }
```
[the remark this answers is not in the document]
deguif: Ok, but just curious what it's done this way? CS?

pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
deguif: So are you sure you want to keep the `empty` check, as `count` should do the same, and the advantage is that it's allowed to initialize the variable at the same time ;)

pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
makasim: which operation are you referring to?

deguif: https://github.com/php-enqueue/enqueue-dev/pull/626/files/f524a099908afa284278ef337b4042339094bf6b#diff-5f17e0abec53c03fec471b2c5f9f668dL34

makasim:
could you please rename timeout argument to `timeoutMs` and use it in `receive` method.

deguif:
Done

pkg/enqueue/Consumption/FallbackSubscriptionConsumer.php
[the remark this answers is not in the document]
deguif: Done

makasim:
Thank you @deguif !

deguif:
You're welcome.

Issue #592, opened 2018-10-29
[bundle][client] Add ability to configure multiple clients. 


makasim:
Bundle already supports it. The configuration should be extended.

makasim:
@drgomesp the feature has just landed to master

drgomesp:
Amazing. Thanks @makasim and @ASKozienko

PR #628, opened 2018-11-09
[bundle] Multi Client Configuration

ASKozienko:
fixes https://github.com/php-enqueue/enqueue-dev/issues/594, https://github.com/php-enqueue/enqueue-dev/issues/592, https://github.com/php-enqueue/enqueue-dev/issues/634

```yaml
enqueue:
   my_amqp: # default
        transport: 'amqp:'
        client:
            traceable_producer: true
        async_commands: true
    my_sqs:
        transport: 'sqs:'
        client: false
        async_commands: false
```

Send a message

```
./bin/console enqueue:produce # use default
./bin/console enqueue:produce --client=my_amqp
./bin/console enqueue:produce --client=my_sqs
```

Consume a message

```
./bin/console enqueue:consume # use default
./bin/console enqueue:consume --client=my_amqp
./bin/console enqueue:consume --client=my_sqs
```

### DIFF
--- a/pkg/enqueue-bundle/DependencyInjection/Configuration.php
+++ b/pkg/enqueue-bundle/DependencyInjection/Configuration.php
@@ -2,6 +2,7 @@
 
 namespace Enqueue\Bundle\DependencyInjection;
 
+use Enqueue\Monitoring\Symfony\DependencyInjection\MonitoringFactory;
 use Enqueue\Symfony\Client\DependencyInjection\ClientFactory;
 use Enqueue\Symfony\DependencyInjection\TransportFactory;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
@@ -26,8 +27,8 @@ final class Configuration implements ConfigurationInterface
             ->always(function ($value) {
                 if (empty($value)) {
                     return [
-                        'transport' => [
-                            'default' => [
+                        'default' => [
+                            'transport' => [
                                 'dsn' => 'null:',
                             ],
                         ],
@@ -36,8 +37,8 @@ final class Configuration implements ConfigurationInterface
 
                 if (is_string($value)) {
                     return [
-                        'transport' => [
-                            'default' => [
+                        'default' => [
+                            'transport' => [
                                 'dsn' => $value,
                             ],
                         ],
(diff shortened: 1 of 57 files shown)


pkg/enqueue-bundle/DependencyInjection/Configuration.php
```diff
@@ -50,57 +50,46 @@ public function getConfigTreeBuilder(): TreeBuilder
...
-        ;
+        $transportConfig = $transportFactory->getConfiguration('transport');
+        $transportConfig->isRequired();
 
-        $transportFactory->addTransportConfiguration($transportPrototypeNode);
+        $consumerConfig = $transportFactory->getQueueConsumerConfiguration('consumption');
 
-        $consumptionNode = $rootNode->children()->arrayNode('consumption');
-        $transportFactory->addQueueConsumerConfiguration($consumptionNode);
+        $clientConfig = (new ClientFactory('default'))->getConfiguration('client', $this->debug);
```
makasim: What if make this getConfiguration methods static? It  is confusing to create a default factory and get a configuration from it. 

@ASKozienko  WDYT?

pkg/enqueue/Symfony/DiUtils.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace Enqueue\Symfony;
+
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\Reference;
+
+class DiUtils
```
makasim: Could you use it everywhere (instead of traits)?

makasim: ~~ingore the comment above~~

pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php
```diff
@@ -25,76 +31,91 @@ public function load(array $configs, ContainerBuilder $container): void
...
-        if ($config['async_commands']['enabled']) {
-            if (false == class_exists(AsyncCommandExtension::class)) {
-                throw new \LogicException('The "enqueue/async-command" package has to be installed.');
-            }
-
-            $extension = new AsyncCommandExtension();
-            $extension->load([[]], $container);
-        }
+        $container->setParameter('enqueue.transports', $transportNames);
+        $container->setParameter('enqueue.clients', $clientNames);
```
makasim: store the name of default transport and client as parameters.

makasim:
atm cli commands assume `default` is default transport\client but it is not a case any more. Commands should be updated.

PR #630, opened 2018-11-12
fixed headline

OskarStark:


makasim:
@OskarStark I suppose it should go to 0.8.

PR #639, opened 2018-11-16
Fix compiler pass

ASKozienko:


pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php
```diff
@@ -43,7 +43,7 @@ public function load(array $configs, ContainerBuilder $container): void
             }
 
             // or with name 'default'
-            if ('default' === $name) {
+            if (DiUtils::DEFAULT_CONFIG === $name) {
```
makasim: not sure we need a constant

makasim: it could confuse people

Issue #638, opened 2018-11-15
[dbal] consumer should still work if table is truncated


makasim:
for more infor: https://github.com/php-enqueue/enqueue-dev/pull/613#discussion_r230595264

PR #640, opened 2018-11-16
[dbal] handle gracefully concurrency issues or 3rd party interruptions. 

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/638

### DIFF
--- a/pkg/dbal/DbalConsumerHelperTrait.php
+++ b/pkg/dbal/DbalConsumerHelperTrait.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Enqueue\Dbal;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Exception\DeadlockException;
 use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Types\Type;
 use Ramsey\Uuid\Uuid;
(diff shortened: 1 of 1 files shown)


pkg/dbal/DbalConsumerHelperTrait.php
```diff
@@ -105,9 +107,13 @@ protected function redeliverMessages(): void
             ->setParameter('redelivered', true, Type::BOOLEAN)
         ;
 
-        $update->execute();
+        try {
+            $update->execute();
 
-        $this->redeliverMessagesLastExecutedAt = microtime(true);
+            $this->redeliverMessagesLastExecutedAt = microtime(true);
+        } catch (DeadlockException $e) {
```
Tobion: it would be better to use `\Doctrine\DBAL\Exception\RetryableException` that I created. this would also cover lock timeouts.

Tobion:
https://github.com/php-enqueue/enqueue-dev/blob/db7c6bd342f4ac113476f18a2744ec07ffe3eab5/pkg/dbal/DbalConsumerHelperTrait.php#L66 can also cause errors likely that you should catch.

makasim:
@Tobion  https://github.com/php-enqueue/enqueue-dev/pull/640

PR #642, opened 2018-11-16
[dbal] Use RetryableException, wrap fetchMessage exception to it too.

makasim:
addition to https://github.com/php-enqueue/enqueue-dev/pull/640


Tobion:
👍

PR #649, opened 2018-11-22
Fix ping to check each connection, not only first one

webmake:


makasim:
@webmake could you change the base branch to 0.8. It would be good to have it fixed there too.

webmake:
It seems random test failure:
```
1) Enqueue\Dbal\Tests\Functional\DbalConsumerTest::testShouldOrderMessagesWithSamePriorityByPublishedAtDate
InvalidArgumentException: $bytes string should contain 16 characters.
2) Enqueue\SimpleClient\Tests\Functional\SimpleClientTest::testSendEventWithTwoSubscriber with data set "dbal_dsn" (array('mysql://root:rootpass@mysql/mqdev'), '+1sec')
InvalidArgumentException: $bytes string should contain 16 characters.
```
not related with a change

makasim:
@webmake yeap.

makasim:
@webmake it added to much an related changes to the PR (because of base branch change). Would it be possible to cherry-pick your commit to a new PR with 0.8 as a base?

webmake:
Yeah, i see. Okay

PR #651, opened 2018-11-22
Fix ping to check each connection, not only first one

webmake:
(cherry picked from commit 72d4097)


webmake:
Reference to original PR https://github.com/php-enqueue/enqueue-dev/pull/649

PR #661, opened 2018-11-28
[dsn] remove commented out code

kunicmarko20:
Was reviewing the dsn class since I plan to use it and I saw some commented code. If this was intentional just close the PR.

Looks like a nice lib, thank you!


makasim:
@kunicmarko20 you are right, those methods should be removed.

PR #689, opened 2018-12-14
updated docs

OskarStark:
hope this is correct @makasim


makasim:
there is a compiler pass that check that everything is in order https://github.com/php-enqueue/enqueue-dev/blob/bede5416ead0dcdae8ba2f6ff93333af115f4a3c/pkg/enqueue/Symfony/Client/DependencyInjection/AnalyzeRouteCollectionPass.php#L50

PR #691, opened 2018-12-14
updated PHPDoc

OskarStark:
> It is possible to pass other options, they could be accessible on a route instance through options.

Is this sentence still valid?


makasim:
> Is this sentence still valid?

Yes, it is

makasim:
Sorry, The `exclusive` option exists and still in use.

PR #709, opened 2018-12-27
[consumption] fix wrong niceness extension param

ramunasd:
Values from CLI comes as strings but extension expects integer, therefore exception is thrown on any value, even correct one.


ramunasd:
@makasim fixed

Issue #475, opened 2018-07-16
[sqs] Dead Letter Queue Adoption


w3evolutions:
When the `QueueConsumer` receives a `self::REJECT` and calls `$consumer->reject($message, false);` on the `SqsConsumer` the message is deleted. This doesn't allow for the SQS Queue to force it's Dead Letter Maximum Receives, as the message is deleted.

Shouldn't there be a way to reject a message but leave it in SQS so the next iteration of `QueueConsumer->consume()` will pull the same message and it's received count will rise and eventually hit the Maximum Receives?

Or am I doing something wrong?

makasim:
> Shouldn't there be a way to reject a message but leave it in SQS so the next iteration of QueueConsumer->consume() will pull the same message and it's received count will rise and eventually hit the Maximum Receives?

I am not aware of such feature. Could you please post a link or code example (maybe with SQS SDK) on how to do that?

w3evolutions:
in enqueue\sqs\SqsConsumer.php line 133+. No matter what the SqsConsumer receives it always calls $this->context->deleteMessage(). Because a message is always deleted SQS will never hit it's dead letter queue. One would need to call:

```php
public function getQueueAttributes()
{
	$this->context->getQueueAttributesAsync([
		'QueueUrl' => $this->queueUrl,
		'AttributeNames' => ['RedrivePolicy']
	])
	->then(function (Result $queueAttributesResult) {
		// Do work here to know if there is a redrive policy
	});
}
```
**enqueue\sqs\SqsConsumer.php**
```php
public function acknowledge(PsrMessage $message)
{
	InvalidMessageException::assertMessageInstanceOf($message, SqsMessage::class);

	$this->context->getClient()->deleteMessage([
		'QueueUrl' => $this->context->getQueueUrl($this->queue),
		'ReceiptHandle' => $message->getReceiptHandle(),
	]);
}

public function reject(PsrMessage $message, $requeue = false)
{
	InvalidMessageException::assertMessageInstanceOf($message, SqsMessage::class);

	$this->context->getClient()->deleteMessage([
		'QueueUrl' => $this->context->getQueueUrl($this->queue),
		'ReceiptHandle' => $message->getReceiptHandle(),
	]);

	if ($requeue) {
		$this->context->createProducer()->send($this->queue, $message);
	}
}
```

makasim:
[...]

PR #720, opened 2019-01-11
[sqs] Dead Letter Queue Adoption

cshum:
Per #475, current Sqs requeue behaviour deletes the message and create another one.

This is not desirable because SQS lost track of Receive count, hence unable to adopt dead letter queue.

Changes made to terminte visibility timeout instead:
https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#terminating-message-visibility-timeout

Todo:
- [x] Fix Tests

### DIFF
--- a/pkg/sqs/SqsClient.php
+++ b/pkg/sqs/SqsClient.php
@@ -43,6 +43,11 @@ class SqsClient
         return $this->callApi('receiveMessage', $args);
     }
 
+    public function changeMessageVisibility(array $args): Result
+    {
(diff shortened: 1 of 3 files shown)


makasim:
Thanks for the contribution @cshum. Could you please fix the failing test https://travis-ci.org/php-enqueue/enqueue-dev/jobs/478393876 ?

cshum:
@makasim unit test fixed. Please review, thanks.

PR #722, opened 2019-01-14
[pheanstalk] Requeuing a message should not acknowledge it beforehand

alanpoulain:
Requeuing a message in Beanstalk means to release a reserved message:
https://github.com/beanstalkd/beanstalkd/blob/b5a6f7a23a368ffb31fbf48cdffe95541166d3fa/doc/protocol.txt#L85-L102
If the message is acknowledged beforehand (i.e. deleted), it can't be released anymore.
If you are OK with the modification, I will add a unit test.


alanpoulain:
No unit test then? :stuck_out_tongue:

makasim:
@alanpoulain ah, missed that. Could you add it?

Issue #731, opened 2019-01-18
[dbal] message is not added to the queue


otzy:
I use dbal package 0.9.6 with mysql.
producer->send() does not add a record to the queue.

$this->context->getDbalConnection()->insert() in DbalProducer.php:118 returns 0 but no any exception thrown.

But if I comment out the third parameter with types, the message is added successfully.
The problem seems to be in the parameter `'redelivered' => Type::BOOLEAN,`
When it's changed to `'redelivered' => Type::SMALLINT,` everything works

otzy:
I think it's also worth to check that insert() returned 1 and throw an exception otherwise.

And btw. it will be nice if you inherited all your exceptions from RuntimeException. Most of them I think caused by the wrong usage of library and there is no need to catch them in the app

makasim:
that strange since there are some functional tests which pass. 

Would you like to submit a PR?

PR #733, opened 2019-01-21
 Exception on affected record !=1

otzy:
additional exception; avoid using Type::BOOLEAN

### DIFF
--- a/pkg/dbal/DbalProducer.php
+++ b/pkg/dbal/DbalProducer.php
@@ -129,7 +129,7 @@ class DbalProducer implements Producer
                 'delivery_id' => Type::STRING,
                 'redeliver_after' => Type::BIGINT,
             ]);
-            
+
             if (1 !== $rowsAffected) {
                 throw new Exception('The message was not enqueued. Dbal did not confirm that the record is inserted.');
             }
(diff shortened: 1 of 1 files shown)


makasim:
could you please fix code style issues reported by travis?

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/731

otzy:
done

pkg/dbal/DbalProducer.php
```diff
@@ -125,10 +125,14 @@ public function send(Destination $destination, Message $message): void
                 'queue' => Type::STRING,
                 'time_to_live' => Type::INTEGER,
                 'delayed_until' => Type::INTEGER,
-                'redelivered' => Type::BOOLEAN,
+                'redelivered' => Type::SMALLINT,
```
makasim: This is strange. Functional tests are working with the current code, and as far as I know doctrine should handle boolean type internally according to platform driver.

makasim: IMO, should be investigated deeper.

otzy: I found people had similar problems with boolean. For eample https://stackoverflow.com/questions/10242312/pdo-bindvalue-with-pdoparam-bool-causes-statement-execute-to-fail-silently

Try to search "insert with PDO::PARAM_BOOL fails"
I have php 7.3 and mysql 5.7, but the issue is still here.

May be if you could use SMALLINT everywhere for the type of this field it could be a solution.

PR #755, opened 2019-02-04
[amqp-lib] supporting channel_rpc_timeout option

derek9gag:
Adding support of the new argument `channel_rpc_timeout` introduced in `php-amqplib` 2.8.2  (https://github.com/php-amqplib/php-amqplib/pull/609). This is a new optional arg in the constructor of `AMQPConnection` so it is backward compatible with `~2.8`.

### DIFF
--- a/pkg/amqp-lib/AmqpConnectionFactory.php
+++ b/pkg/amqp-lib/AmqpConnectionFactory.php
@@ -47,6 +47,7 @@ class AmqpConnectionFactory implements InteropAmqpConnectionFactory, DelayStrate
             ->addDefaultOption('login_response', null)
             ->addDefaultOption('locale', 'en_US')
             ->addDefaultOption('keepalive', false)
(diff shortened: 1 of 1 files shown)


pkg/amqp-tools/ConnectionConfig.php
```diff
@@ -226,29 +229,37 @@ public function getVHost()
...
      */
     public function getConnectionTimeout()
     {
         return $this->getOption('connection_timeout');
     }
 
+    /**
+     * @return float
+     */
+    public function getChannelRpcTimeout()
```
makasim: This is a php-amqplib specific option, I don't think we should introduce a dedicated getter for it in amqp tools. 

Could you use getOption method to get its value ?

derekchan: Updated to limit code changes to `pkg/amqp-lib`.

makasim:
@derek9gag ready to merge?

Issue #750, opened 2019-02-01
[Redis] messages lost and non-atomic operations


cshum:
During our usage of Enqueue Redis on a  possibly unstable connections with lots of retries, we occurred ~3-5 message lost out of ~50000 messages. 

Having looked into source the [receiveMessage/processResult](https://github.com/php-enqueue/redis/blob/master/RedisConsumerHelperTrait.php#L23) [acknowledge/requeue](https://github.com/php-enqueue/redis/blob/master/RedisConsumer.php#L114) are indeed non-atomic operations like brpop + zadd, zrem + lpush etc.

Do you think if implementations of atomic operations, like [RPOPLPUSH](https://redis.io/commands/rpoplpush), [BRPOPLPUSH](https://redis.io/commands/brpoplpush), or any kind of multi or lua scripts would be feasible?

makasim:
That should be addressed indeed, not sure what would be the best approach to fix it though.

cshum:
Atomic ones are probably harder to implement.

What about setting add operations _before_ the remove ones? There may be duplicated messages but still better than message lost.

makasim:
If you need strict message delivery guarantee than I'd suggest you to use a real broker such as RabbitMQ or AmazonSQS.

We've investigated the issue at our end looking for a solution to fix it. Unfortunately, we've not come with any better approach. The current one is not ideal but it remains the best one

PR #757, opened 2019-02-05
[Redis] message lost prevention

cshum:
Per #750, the main idea of changes is to swap BRPOP/RPOP into BRPOPLPUSH/RPOPLPUSH and make sure writes happen before deletes:

* On receiveMessage, BRPOPLPUSH/RPOPLPUSH fetches result, also move it into `:processing` list
* At the end processResult after ZADD, LREM to remove `:processing` messages
* If worker crashed/connection drops between receiveMessage and processResult, messages will still left intact under the `:processing` list
* When all jobs clear and queue empty, RENAMENX to migrate `:processing` list into the main queue.

There are still stuffs need to be done. But the current version is running on our production, and seems to be working so far.
With the revised design, tradeoff would be message duplications. If worker jobs are idempotent then this will not be a concern. 
To cope with duplicated issue, can keep a hash set of uuids and check against it. The hash set can be deleted when queue becomes empty. 

Plans/Todo:
- [x] redis interface add methods
  -   brpoplpush
  -   rpoplpush
  -   lrem
  -   renamenx
- [x] predis driver added methods
- [ ] phpredis driver added methods
- [x] disable RedisSubscriptionConsumer, as BRPOPLPUSH does not support multiple queues, not sure how to fix this.
- [x] Refractor receiveMessage:
  -   brpoplpush/rpoplpush from queue to `:processing`
  -   zadd reserved
  -   lrem `queue:processing`
- [x] Refractor requeue:
  - swap LPUSH to be called before ZREM
- [x] on queue empty, renamenx `:processing` back to main queue
- [ ] messages duplicates detection (using hash set of uuids)
- [ ] on queue empty, clear duplication hash set
- [ ] Fix/add test cases

### DIFF
--- a/pkg/redis/PRedis.php
+++ b/pkg/redis/PRedis.php
@@ -105,7 +105,7 @@ class PRedis implements Redis
         try {
             return $this->redis->renamenx($key, $target);
         } catch (PRedisServerException $e) {
-            throw new ServerException('lrem command has failed', 0, $e);
+            throw new ServerException('renamenx command has failed', 0, $e);
         }
     }
 
--- a/pkg/redis/PRedis.php
+++ b/pkg/redis/PRedis.php
@@ -104,10 +104,36 @@ class PRedis implements Redis
         }
     }
 
+    public function brpoplpush(string $source, string $dest, int $timeout): ?RedisResult
+    {
+        try {
+            if ($result = $this->redis->brpoplpush($source, $dest, $timeout)) {
+                return new RedisResult($result[0], $result[1]);
+            }
+
+            return null;
+        } catch (PRedisServerException $e) {
+            throw new ServerException('brpop command has failed', 0, $e);
+        }
+    }
+
     public function rpop(string $key): ?RedisResult
     {
         try {
             if ($message = $this->redis->rpop($key)) {
+                return new RedisResult($key, $message);
+            }
+
+            return null;
+        } catch (PRedisServerException $e) {
+            throw new ServerException('rpop command has failed', 0, $e);
+        }
+    }
+
+    public function rpoplpush(string $source, string $dest): ?RedisResult
+    {
+        try {
+            if ($message = $this->redis->rpoplpush($source, $dest)) {
                 return new RedisResult($key, $message);
             }
 
--- a/pkg/redis/PRedis.php
+++ b/pkg/redis/PRedis.php
@@ -91,6 +91,15 @@ class PRedis implements Redis
         }
     }
 
+    public function lrem(string $key, int $count, string $value): int
+    {
+        try {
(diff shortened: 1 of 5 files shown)


makasim:
This could not be merged since it introduces even worse scenarios. With this change, it would not work with more than one consumer at all. 

for example [here](https://github.com/php-enqueue/enqueue-dev/pull/757/files#diff-6620761ad724c4777739a4845bb25fafR27) you migrate all message and that would work for one consumer but fail for more.

makasim:
If you need strict message delivery guarnty than I'd suggest you to use real broker such as RabbitMQ or AmazonSQS.

cshum:
> This could not be merged since it introduces even worse scenarios. With this change, it would not work with more than one consumer at all.
> 
> for example [here](https://github.com/php-enqueue/enqueue-dev/pull/757/files#diff-6620761ad724c4777739a4845bb25fafR27) you migrate all message and that would work for one consumer but fail for more.

https://redis.io/commands/renamenx would work for multiple consumer isn't it? If 1 consumer successful migrate messages, other consumers trying to migrate would result ` returns an error when key does not exist` or `0 if newkey already exists.`.

With the added logic its probably slower and have duplicated messages. In our production we have been running this for a while with ~15 workers, no message lost so far (given redis is not down)

cshum:
> If you need strict message delivery guarantee than I'd suggest you to use a real broker such as RabbitMQ or AmazonSQS.
> 
> We've investigated the issue at our end looking for a solution to fix it. Unfortunately, we've not come with any better approach. The current one is not ideal but it remains the best one

I would suggest dropping this library altogether for next project. 

When someone picks a message queue library they are probably expecting at least one delivery. Any pitfalls should have been mentioned in readme and drop the “battle tested, production ready” bs. Reliability doesn't seems to be you guys concern.

makasim:
Just a friendly reminder, you got a couple of years of several developers work for free

makasim:
and I have to spend my free unpaid time to review your PR

makasim:
You could release your version as a standalone package. I would gladly accept a PR for the doc that describes current implement limitations and possible message lost. It could contain a link to your package too. 

If time proves your approach is better it could be merged here. 

For now, I don't have enough resources to go through this PR and be sure it does not break anything

cshum:
Thanks. The PR is clearly unfinished and tests are failed, of course I am not expecting this to be merged any time soon. Neither do I have time to polish it good enough for general release. 

We had problems and we have attempted to fix it. I just though it was a good idea to share the progress though. The source here is exactly the same as what we are currently running.

makasim:
@cshum fyi https://github.com/php-enqueue/enqueue-dev/pull/770

PR #764, opened 2019-02-14
[fs] polling_interval config should be milliseconds not microseconds

ssiergl:
regression from this commit https://github.com/php-enqueue/fs/commit/c5199ca5bac5c661aed86d6b2f751630af42378d

config should be milliseconds like in docs stated.....$this->pollingInterval is holding milliseconds which is correct.

But usleep is using microseconds. To fix it just convert milliseconds to microseconds


ssiergl:
@makasim

PR #765, opened 2019-02-14
[dbal]: make dbal connection config usable again

ssiergl:
right now user cannot declare connection options.

Reason is that always 'connection' => 'url' was set which dbal always prefers over other
connection settings. With this fix connection can be defined dbal compliant like this:

      connection:
        host: 'localhost'
        port: '3306'
        dbname: 'app'
        user: 'app'
        password: ''

'dsn only' works like before.

@makasim what do you think?

### DIFF
--- a/pkg/dbal/Tests/DbalConnectionFactoryConfigTest.php
+++ b/pkg/dbal/Tests/DbalConnectionFactoryConfigTest.php
@@ -102,6 +102,72 @@ class DbalConnectionFactoryConfigTest extends TestCase
         ];
 
         yield [
+            [
+                'dsn' => 'mysql+pdo:',
+                'connection' => [
+                    'dbname' => 'customDbName',
+                ],
+            ],
+            [
+                'connection' => [
+                    'dbname' => 'customDbName',
+                    'driver' => 'pdo_mysql',
+                    'host' => 'localhost',
+                    'port' => '3306',
+                    'user' => 'root',
+                    'password' => '',
+                ],
+                'table_name' => 'enqueue',
+                'polling_interval' => 1000,
+                'lazy' => true,
+            ],
+        ];
+
+        yield [
+            [
+                'dsn' => 'mysql+pdo:',
+                'connection' => [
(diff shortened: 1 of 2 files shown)


makasim:
This is nice, could you please add a test for this case to DbalConnectionFactoryConfigTest?

ssiergl:
yeah right now working on a pimcore plugin and the connection variables are pretty long...so the dsn is nearly unreadable :-p

added config tests. Not tested...but should be ok

pkg/dbal/Tests/DbalConnectionFactoryConfigTest.php
```diff
@@ -101,6 +101,54 @@ public static function provideConfigs()
...
+                ],
+                'table_name' => 'enqueue',
+                'polling_interval' => 1000,
+                'lazy' => true,
+            ],
+        ];
+
+        yield [
+            [
+                'dsn' => 'mysql+pdo://user1:pass1@host1:10001/db',
```
makasim: parameters from dsn should overwrite ones set in the configuration.

ssiergl: Done. Due that dbal ignores other connection settings if `url` is set we simply just set the fq `url` and we are done.

ssiergl:
tested all possible configurations...looks all well

pkg/dbal/Tests/DbalConnectionFactoryConfigTest.php
```diff
@@ -101,6 +101,72 @@ public static function provideConfigs()
             ],
         ];
 
+        yield [
+            [
+                'dsn' => 'mysql+pdo:',
+                'connection' => [
+                    'dbname' => 'dbname',
```
makasim: could you please use `customDbName` to reflect that it overwrites default value.

ssiergl: Done

PR #768, opened 2019-02-16
Allow to change visibility timeout on SQS during requeue message

mmp4k:
I think, hardcoding "0" during requeue is not the best option. Especially then `SqsMessage` already has method `getDelaySeconds`.


makasim:
This not correct, requeue means the message should be redelivered right away.

deguif:
Just seeing this issue, I just make the same comment here: https://github.com/php-enqueue/enqueue-dev/commit/257fdb5fbcb95d141ced3de2fc0d1dfd2a53fb3d#r33577211

And this solution but which doesn't work for SQS as the message is not repushed.

PR #771, opened 2019-02-20
Catch signal in Bunny adapter

snapshotpl:
Bunny adapter try to catch `ClientException` when signal was triggered and exception's message is `stream_select() failed.`. However message is `stream_select() failed: Unknown error.`.


pkg/amqp-bunny/AmqpSubscriptionConsumer.php
```diff
@@ -47,7 +47,7 @@ public function consume(int $timeout = 0): void
         try {
             $this->context->getBunnyChannel()->getClient()->run(0 !== $timeout ? $timeout / 1000 : null);
         } catch (ClientException $e) {
-            if ('stream_select() failed.' == $e->getMessage() && $signalHandler->wasThereSignal()) {
+            if ('stream_select() failed' === substr($e->getMessage(), 0, 22) && $signalHandler->wasThereSignal()) {
```
makasim: would it be better to do `0 === strpos` to make sure it is the beginning of the exception message.

Issue #773, opened 2019-02-21
[bundle][asyncEvents] Reenqueue on failure


nataliaAznar:
I am trying to reenqueue an event when it fails, but it is dispatched as a sync event.
Is there any way to reenqueue an event when it fails?

makasim:
[Sync mode](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/async-event-dispatcher/AsyncListener.php#L58) has to be disabled for the event you want to requeue. 

There is no way to do it for a single event atm

nataliaAznar:
I have already checked that.
My problem is that, in the event, although  it is defined as asynchronous, the sync mode has not been restored when it is executed. Because of this, when I requeue the event, it is dispached as a sync event.
I am trying to inject the AsyncListener in the event, but it seems that is is not defined when de listnere is called:

Cannot autowire service "Listener": argument "$listener" of method "__construct()" references class "Enqueue\AsyncEventDispatcher\AsyncListener" but no such service exists. You should maybe alias this class to the existing "enqueue.events.async_listener" service.

makasim:
https://github.com/php-enqueue/enqueue-dev/pull/777#issuecomment-492535856

PR #777, opened 2019-02-21
Allow to set an Event as async

nataliaAznar:
This fix issue #773

Allow to set an Event as async, in order to being able to requeue it

### DIFF
--- a/pkg/async-event-dispatcher/AsyncListener.php
+++ b/pkg/async-event-dispatcher/AsyncListener.php
@@ -59,6 +59,16 @@ class AsyncListener
     }
 
     /**
+     * @param $eventName
+     */
+    public function asyncMode($eventName)
+    {
+        if($this->isSyncMode($eventName)) {
+            unset($this->syncMode[$eventName]);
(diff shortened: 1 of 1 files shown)


makasim:
This might have some unwanted side effects. 

I'd better move [this code](memo_bot) to separate public method and call it for the even you want requeue.

nataliaAznar:
But that means that you cannot use the same code to enqueue and requeue an Event. It forces you to have duplicated code and evaluating if you are enqueueing an Event for the first time or not. Is there any way to check this?
It also means that you must know if an Event is sync or async, and the code must change depending on the type of the Event. The other way, you can unset the Event from the sync array and, if it is not async, it will still be executed on the rigth way

makasim:
My gut feeling tells me that this change could lead to a very serious bug such as endless cycles. What if an event you process triggers same async event that goes to queue and then picked up by a processor that unsets sync mode and there is no end for it. 

We should find a solution to address the use case you have but it should be carefully thought out.

PR #785, opened 2019-02-26
Fix amqp-tools dependency

### LINKED ISSUE #783: Doc example doesn't work
In the documentation it's written that you can instanciate a factory (for localhost) with : 

`$factory = new \Enqueue\AmqpLib\AmqpConnectionFactory();`

Though i get an error stating that the Enqueue\Dsn\Dsn class cannot be found. Everythink works fine if i give the factory an empty array.

https://github.com/php-enqueue/enqueue-dev/blob/master/docs/transport/amqp_bunny.md

TomPradat:
Fix #783 

The enqueue/dsn must be required and not only for dev because ConnectionConfig
uses it.

### DIFF
--- a/pkg/amqp-tools/composer.json
+++ b/pkg/amqp-tools/composer.json
@@ -8,13 +8,13 @@
     "require": {
         "php": "^7.1.3",
         "queue-interop/amqp-interop": "^0.8",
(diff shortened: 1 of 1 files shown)


makasim:
Thanks!

PR #787, opened 2019-02-26
Add upgrade instructions

KDederichs:


makasim:
@KDederichs thank you!

Issue #749, opened 2019-02-01
[kafka] RdKafkaProducer has no way of error handling


Steveb-p:
When an error occurs in `RdKafkaProducer`, it is silently ignored - or, more accurately, since it occurs outside of the main thread of php/php-fpm, it is not reported back.

A simple test can be done to see this issue: trying to send a message to a non-existent Kafka server will result in nothing (see point 2. though)

This results in two things:
1. Process has no way of knowing, that a message that was supposed to be delivered will be not. Producer returns immediately without waiting for message to be acknowledged.
2. Due to how it is handled in arnaud-lb/php-rdkafka, process that was supposed to send the message will be locked and retry the operation for a long time (from my testing it seems around 5 mins, wasn't able to find a configuration option to change it to something else). In my case with default configuration of php-fpm docker image it resulted in fpm pool becoming locked after 5 requests being made, since thats the default configuration for max spawned children of php-fpm.  
While this particular part is not really possible to fix inside enqueue, it's important since the message might actually be delivered later on.

I'd like to ask for opinion regarding how RdKafkaProducer should handle this situation. IMO it is worthwile to add a configuration option to make sending messages synchronous for this particular Producer - or at least wait a specified amount of time for message to be potentially acknowledged.
This can be done by introducing this code at the end of `send` method:
```php
$topic = $this->producer->newTopic($destination->getTopicName(), $destination->getConf());
$topic->produce($partition, 0 /* must be 0 */, $payload, $key);
$start = microtime(true);
while ($this->producer->getOutQLen() > 0) {
     $this->producer->poll(1);

     if (microtime(true) - $start > 10) {
           throw new \RuntimeException("Message sending failed");
     }
}
```

This has a side effect of actually *calling `dr_msg_cb` & `error_cb` callbacks*, which are otherwise ignored (or at least that's what my testing indicated).

Thoughts?

makasim:
I don't have an immediate answer. Will think about it

Steveb-p:
When callbacks are set up like so:
```php
echo microtime(true), '<br />';
$this->conf->setErrorCb(function ($kafka, $err, $reason) {
    echo nl2br(sprintf("Kafka error: %s (reason: %s)\n", rd_kafka_err2str($err), $reason));
    echo microtime(true), '<br />';
});

$this->conf->setDrMsgCb(function ($kafka, $message) {
    if ($message->err) {
        // message permanently failed to be delivered
[...]

PR #798, opened 2019-03-13
stats_cb support

fkulakov:
Hello. I wants to support stats_cb config field first of all for my own needs but for overall system flexibility too.

### DIFF
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -186,6 +186,10 @@ class RdKafkaContext implements Context
                 $this->conf->setRebalanceCb($this->config['rebalance_cb']);
             }
 
+            if (isset($this->config['stats_cb'])) {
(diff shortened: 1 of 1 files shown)


Steveb-p:
While configuration-wise it should work, it will probably fail due to https://github.com/php-enqueue/enqueue-dev/issues/749. Specifically because `librdkafka::poll` is never called internally, which also causes any callbacks to **not** fire.

I'll look into fixing the abovementioned issue tonight, if that's alright for you @fkulakov.

See: https://github.com/edenhill/librdkafka/wiki/Consumer-lag-monitoring

> librdkafka automatically monitors consumer lag for RD_KAFKA_CONSUMER handles and the information is available in the statistics JSON object as emitted by stats_cb. A statistics callback is set up with rd_kafka_conf_set_stats_cb() and the callback is **called at the configured interval (statistics.interval.ms) from an rd_kafka_poll() call**.

Steveb-p:
Totally forgot :)
Thanks for contributing @fkulakov

PR #809, opened 2019-03-28
[InfluxDB][Monitoring] Allow passing Client as configuration option.

### LINKED ISSUE #803: [monitoring][influxdb] Do not attempt to create database / Allow TCP socket / custom Client
I've created a PR for influxdb-php library introducing ability to write to TCP sockets, for example telegraf's ones - which allows me to write to a local socket instead of directly writing measurements to influxdb. (influxdata/influxdb-php#124)
I'm using a custom version of InfluxDBStorage class (monitoring) because the one from enqueue/monitoring always does try to create a database...
https://github.com/php-enqueue/enqueue-dev/blob/0652c5c987395ad3af1f226147f0f8c7f1c6540f/pkg/monitoring/InfluxDbStorage.php#L164-L181
...which won't work in above case, because that socket is unable to handle database queries. I'm passing a custom client directly (as constructor `$config` variable)...
https://github.com/php-enqueue/enqueue-dev/blob/0652c5c987395ad3af1f226147f0f8c7f1c6540f/pkg/monitoring/InfluxDbStorage.php#L49-L80
... but it appears that there is no way to pass configuration for measurements writing (since you can either pass a Client instance or config string/array).

I believe there should be a setter for configuration for this case? WDYT? Should using `Client` or `Database` instance be configurable as well / instead?

EDIT: For writing I'm replacing
https://github.com/php-enqueue/enqueue-dev/blob/0652c5c987395ad3af1f226147f0f8c7f1c6540f/pkg/monitoring/InfluxDbStorage.php#L112
with
```
$this->client->write([], $points);
```
since the currently used in php-enqueue code will call `Database`'s
https://github.com/influxdata/influxdb-php/blob/4a1efb43656a4f2b390201865cfe7051c895dff7/src/InfluxDB/Database.php#L162-L179
eventually anyway with passed `Points` as payload.

Steveb-p:
also prevent calling database creation if Client driver does not support QueryDriverInterface.

This is a follow up from a discussion regarding #803 and a quick talk on gitter

> @steveb_gitlab I've not digged much into the issue. I guess passing a client as a part of config should work out. wdyt?

There are a couple of comments left that I should probably drop, but I've included them to make it clear for you why the code looks in a particular way (regarding `doWrite` method).  
Please tell me if there are any code style changes that you want (if you don't want `sprintf` calls maybe?) or anything else.

Fixes #803.

### DIFF
--- a/pkg/monitoring/InfluxDbStorage.php
+++ b/pkg/monitoring/InfluxDbStorage.php
@@ -59,10 +59,17 @@ class InfluxDbStorage implements StatsStorage
         if (empty($config)) {
             $config = [];
         } elseif (is_string($config)) {
-            $config = $this->parseDsn($config);
+            $config = self::parseDsn($config);
         } elseif (is_array($config)) {
-            $config = empty($config['dsn']) ? $config : $this->parseDsn($config['dsn']);
+            $config = empty($config['dsn']) ? $config : self::parseDsn($config['dsn']);
         } elseif ($config instanceof Client) {
+            // Passing Client instead of array config is deprecated because it prevents setting any configuration values
+            // and causes library to use defaults.
+            @trigger_error(
+                sprintf('Passing %s as %s argument is deprecated. Pass it as "client" array property or use createWithClient instead',
+                Client::class,
+                __METHOD__
+            ), E_USER_DEPRECATED);
             $this->client = $config;
             $config = [];
         } else {
@@ -95,6 +102,22 @@ class InfluxDbStorage implements StatsStorage
         }
 
         $this->config = $config;
+    }
+
+    /**
+     * @param Client $client
+     * @param string $config
+     *
+     * @return InfluxDbStorage
+     */
+    public static function createWithClient(Client $client, $config = 'influxdb:'): self
+    {
+        if (is_string($config)) {
+            $config = self::parseDsn($config);
+        }
+        $config['client'] = $client;
+
+        return new static($config);
     }
 
     public function pushConsumerStats(ConsumerStats $stats): void
@@ -221,7 +244,7 @@ class InfluxDbStorage implements StatsStorage
         return $this->database;
     }
 
-    private function parseDsn(string $dsn): array
+    private static function parseDsn(string $dsn): array
     {
         $dsn = Dsn::parseFirst($dsn);
 
--- a/pkg/monitoring/InfluxDbStorage.php
+++ b/pkg/monitoring/InfluxDbStorage.php
@@ -204,8 +204,22 @@ class InfluxDbStorage implements StatsStorage
 
     private function doWrite(array $points): void
     {
-        if (!$this->client || $this->client->getDriver() instanceof QueryDriverInterface) {
-            $this->getDb()->writePoints($points, Database::PRECISION_MILLISECONDS, $this->config['retentionPolicy']);
+        if (null === $this->client) {
+            $this->client = new Client(
+                $this->config['host'],
+                $this->config['port'],
+                $this->config['user'],
+                $this->config['password']
+            );
+        }
+
+        if ($this->client->getDriver() instanceof QueryDriverInterface) {
+            if (null === $this->database) {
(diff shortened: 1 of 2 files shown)


pkg/monitoring/InfluxDbStorage.php
```diff
@@ -158,21 +183,44 @@ public function pushSentMessageStats(SentMessageStats $stats): void
...
-                $this->client = new Client(
-                    $this->config['host'],
-                    $this->config['port'],
-                    $this->config['user'],
-                    $this->config['password']
-                );
+        if (!$this->client || $this->client->getDriver() instanceof QueryDriverInterface) {
+            $this->getDb()->writePoints($points, Database::PRECISION_MILLISECONDS, $this->config['retentionPolicy']);
+        } else {
+            // Code below mirrors what `writePoints` method of Database does.
```
Steveb-p: I was wondering if maybe we should just write to `Client` always instead. 

I'll probably propose to move `writePayload` method into `Client` instance in `influxdb-php`, since `Client` is called anyway and overall `Client` should know how to handle it's parameters when writing.

makasim: Should we wait with this PR it a PR to influxdb-php is merged?

Steveb-p: Honestly influxdb-php maintainers respond kinda slow and it will probably take some time before any such change is merged into it :)
My original PR for it (with TCP driver) does not change the API in any way, and the one I've suggested above will change their API (adding a new method to `Client`). So I'd say don't wait for it?
If such a change is introduced to influxdb-php it will mean that php-enqueue will have either maintain this fallback for a while or only allow influxdb-php versions that contain the fix.

makasim: I am fine to get it merged, one question: do we really need to deprecate passing a client as an argument. As for me, it still a valid option.

Steveb-p: I'd suggest deprecating it, because it basically prevents setting any other options, like database name, `measurementSentMessages`, `measurementConsumedMessages` and `measurementConsumers` (others being basically configuration options for the client itself).

In my opinion it would be best to be as strict as possible when creating function definitions, in this case a constructor. If there is a secondary constructor needed, it can always take the form of a static function, like so:
```php
public static function createFromClient(Client $client, array $config): self
{
    $self = new static($config);
    $self->client = $client;
    return $self;
}
```
or similar. This would allow us to skip most of type checks, as PHP will do a function argument check on it's own and function declaration would explicitly say what value types are acceptable.

I'll drop the deprecation for now anyway.

makasim: okay, let's do it.

Steveb-p: Should I revert the last commit then? :)

Steveb-p: I'll make a part of another PR maybe, so this could be merged and anything else regarding constructor parameters and so on can be left for discussion.

makasim: You can add a static create method as part of this PR and let's deprecate client arg. No need for extra PR.

Steveb-p: Commited. 

I've taken the liberty to convert `parseDsn` method into a static function, since it did not make use of `$this`, which allowed me to use it inside `createWithClient` static function.

I've noticed that there is a possibility of array configuration values to become overwritten, if `dsn` array element is passed, but I'm leaving it as is. From what I understand, it's possible that Symfony bundle might be passing `$config` with `dsn` field which would break if removed (?).

Otherwise I think it's good to go, unless you want some other changes as well?

Steveb-p:
Since there are not further change requests, can I merge it myself @makasim ?

(Since it's outside of my usual scope - Kafka - I don't want to impose myself)

makasim:
Yeap! Once approved you can merge PR. I don't do it because there might be something you want to flesh up.

PR #815, opened 2019-04-01
[dbal] Fix DBAL Consumer duplicating messages when rejecting with requeue

### LINKED ISSUE #814: [dbal] requeue leads to duplicate entries
I Just tried to implement a simple JSON submit queue where the endpoint can either respond HTTP status code 200 or fail. For that fail case I wanted to use the requeue option.

Building a very basic approach I always end up with endless new entries in the enqueue table, every retry leading to a new entry. Maybe I'm totally reading the docs wrong:

`Interop\Queue\Processor::REQUEUE` defines itself as `The original message is removed from the queue but a copy is published to the queue again.`. Looking at the code this is exact scenario is not reached, as the function returns prior removing the old message.

https://github.com/php-enqueue/enqueue-dev/blob/947eb72a5cb1139fcdf4b0fda899db8221ba1d13/pkg/dbal/DbalConsumer.php#L112

In addition to this the `Enqueue\Dbal\DbalProducer::send()` never looks for duplicates and updates them, instead always creating a new entry. This also leads to an infinite loop that fills the database with new entries which are directly processed in the same run.

Is this intended behaviour?

Steveb-p:
Fix #814 

I've switched to `acknowledge` first and then proceed with queueing the message again since this seems how other (Redis, RdKafka for example) work. Others seem to assume that message is automatically acknowledged when it's received.

For the same reason I'm using `acknowledge` instead of `deleteMessage` (other packages seem to work this way.

### DIFF
--- a/pkg/dbal/DbalConsumer.php
+++ b/pkg/dbal/DbalConsumer.php
@@ -100,16 +100,14 @@ class DbalConsumer implements Consumer
     {
         InvalidMessageException::assertMessageInstanceOf($message, DbalMessage::class);
 
+        $this->acknowledge($message);
+
         if ($requeue) {
             $message = clone $message;
             $message->setRedelivered(false);
 
             $this->getContext()->createProducer()->send($this->queue, $message);
-
-            return;
         }
-
-        $this->deleteMessage($message->getDeliveryId());
     }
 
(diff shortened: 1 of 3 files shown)


makasim:
Thank you @Steveb-p

Issue #719, opened 2019-01-07
Different amqp handle props in different manner.


marhub:
Actual result:
in amqp bunny transport props are put into **headers.application_headers**
in ampq-lib transport, props are put into **headers**

Expected result:
I can see props next to headers, not injected into headers key.

If that's expected result - then someone can please explain me why?

makasim:
This behavior is expected. AMQP sends properties as a special header called "application_headers".

marhub:
Sorry, but you do not get me.

considering following code: 

```php 
$headers = [
       'reply_to' => $replyToQueue->getQueueName(),
       'correlation_id' => Uuid::uuid4()->toString()
];
$message = $this->context->createMessage(
       json_encode($data),
       [], // props
       $headers 
);
```
 when using E**nqueue\AmqpBunny\AmqpConnectionFactory** => message ( seems to be not ok ):

![image](https://user-images.githubusercontent.com/2960432/52218965-feb7a180-289b-11e9-8e66-1e82de78610c.png)

when using  **Enqueue\AmqpLib\AmqpConnectionFactory** => message is different ( seems to be ok )

![image](https://user-images.githubusercontent.com/2960432/52219065-36264e00-289c-11e9-9121-98e11ff74dd5.png)


**I expect them to be same in all implementations!**

makasim:
Agree, have you found out why this happens?

marhub:
I belive it's due to internal implementation of bunny lib

makasim:
> I belive it's due to internal implementation of bunny lib

closing than, should be addressed in bunny repo.

PR #816, opened 2019-04-04
Fix bunny producer to properly map headers to expected by bunny headers

nightlinus:
As of [bunny publish method](https://github.com/jakubkulhan/bunny/blob/45702ebe4ef8d5099b5c00f0df71f4b22312203f/src/Bunny/ClientMethods.php#L1769) and [bunny content header frame](https://github.com/jakubkulhan/bunny/blob/45702ebe4ef8d5099b5c00f0df71f4b22312203f/src/Bunny/Protocol/ContentHeaderFrame.php#L120) we need to convert underscore based headers (e.g. delivery_mode) to hyphen based headers (e.g. delivery-mode) when publishing and back when recieving messages. This PR add this convertion back and forward.

Also found old closed issue that was caused by this behaviour #719

### DIFF
--- a/pkg/amqp-bunny/AmqpContext.php
+++ b/pkg/amqp-bunny/AmqpContext.php
@@ -312,6 +312,7 @@ class AmqpContext implements InteropAmqpContext, DelayStrategyAware
     public function convertMessage(BunnyMessage $bunnyMessage): InteropAmqpMessage
     {
         $headers = $bunnyMessage->headers;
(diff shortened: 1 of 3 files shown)


makasim:
@nightlinus It looks good. Could you please make it green so we are good to go.

nightlinus:
Done, last error related to redis. I believe it cant be broken by my pr

PR #818, opened 2019-04-08
Typo on the tag

appeltaert:


Steveb-p:
Whoopsie. Indeed it should be `enqueue.client.extension`.

Thanks @appeltaert

PR #824, opened 2019-04-17
Fixed rabbitmq plugin delay strategy by applying correct header

31vi5:
The delay strategy creates a header array of application_headers which contained the x-delay instead of having the x-delay header directly. Screenshot of the message in the queue https://i.imgur.com/Qk4jyc4.png

The result was that the message was not correctly delayed. I've added the correct header to the message and then the delay started working. I've left the original property unchanged other than removing unnecessary type casting.

### DIFF
--- a/pkg/amqp-tools/RabbitMqDelayPluginDelayStrategy.php
+++ b/pkg/amqp-tools/RabbitMqDelayPluginDelayStrategy.php
@@ -17,7 +17,8 @@ class RabbitMqDelayPluginDelayStrategy implements DelayStrategy
     public function delayMessage(AmqpContext $context, AmqpDestination $dest, AmqpMessage $message, int $delay): void
     {
         $delayMessage = $context->createMessage($message->getBody(), $message->getProperties(), $message->getHeaders());
-        $delayMessage->setProperty('x-delay', (int) $delay);
+        $delayMessage->setProperty('x-delay', $delay);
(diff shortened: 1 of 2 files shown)


makasim:
There are functional tests that cover the use of the plugin ([1](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-lib/Tests/Spec/AmqpSendAndReceiveDelayedMessageWithDelayPluginStrategyTest.php), [2](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-ext/Tests/Spec/AmqpSendAndReceiveDelayedMessageWithDelayPluginStrategyTest.php), [3](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/amqp-bunny/Tests/Spec/AmqpSendAndReceiveDelayedMessageWithDelayPluginStrategyTest.php). They pass.

The issue should be investigated future.  Maybe the tests are wrong, or your change is not relevant.

31vi5:
Perhaps it might have been my misunderstanding of the feature. If I understand it correctly (which now differs from my understanding of the behavior at the time of posting the pull request), the message is sent to queue and then processed right away and then requeued with the delay, is that correct?

makasim:
It depends on a lot of factors. Let me know what you are doing and what you use and what are the versions.

PR #825, opened 2019-04-19
added getQueueLength method to Redis Consumer

13DaGGeR:
Added ability to get count of messages in the queue.
My use case is to log queue size on every message consume.


makasim:
The issue is out of the library scope and could be easily done without any additional code by a consumer extension that takes a Redis client out of the Redis context and performs desired request.

PR #826, opened 2019-04-19
 Add support for using the /topic prefix instead of /exchange.

alessandroniciforo:
Add support for using the /topic prefix instead of /exchange. RabbitMQ needs the /exchange prefix for publishing on an exchange, while ActiveMQ needs the /topic prefix for publishing on a topic

### DIFF
--- a/docs/transport/stomp.md
+++ b/docs/transport/stomp.md
@@ -36,6 +36,12 @@ $factory = new StompConnectionFactory('stomp:');
 // same as above
 $factory = new StompConnectionFactory([]);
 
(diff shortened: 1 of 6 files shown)


makasim:
This is nice. Would you add a scheme extension too? So if `stomp+activemq://` scheme passed it automatically use exchange instead of the topic.

alessandroniciforo:
That's a great idea. I pushed a commit implementing it. The scheme extensions accepted are `rabbitmq` and `activemq`. If omitted, `rabbitmq` is the default for backward compatibility.

PR #845, opened 2019-05-14
[client] Lazy producer.

makasim:
Problem: The proper `ENQUEUE_DSN` is not available during warmup which happens on CI. Warmup would throw an exception if the producer is used in an event listener. To workaround, we should fake DSN. 

The PR introduces a lazy producer which could be safely injected to event listeners and be used during warm up. It throws an exception on `sendEvent` or `sendCommand` call.


ivoba:
Since update to 0.9.10 i get this error in tests, i think its related to this PR:

> Error: Call to undefined method Enqueue\Symfony\Client\LazyProducer::clearTraces()

In test env enqueue is configured to use traceable producer
```

enqueue:
    default:
        transport: '%env(resolve:ENQUEUE_DSN)%'
        client:
            traceable_producer: true
```

In tests clearTraces is used in setUp

```
    protected function setUp()
    {
        parent::setUp();
        /** @var TraceableProducer $producer */
        $producer = static::$container->get(ProducerInterface::class);
        $producer->clearTraces();
    }
```

makasim:
this works for me:

```php
<?php
    private function getProducer(): TraceableProducer
    {
        return self::$container->get('enqueue.client.default.traceable_producer');
    }

```

PR #846, opened 2019-05-15
Message serialization. 

makasim:
needs https://github.com/queue-interop/queue-interop/pull/28

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,8 +11,8 @@
     "ext-mongodb": "^1.3",
     "ext-rdkafka": "^3.0.3",
 
-    "queue-interop/amqp-interop": "^0.8",
-    "queue-interop/queue-interop": "^0.7|^0.8",
+    "queue-interop/amqp-interop": "dev-master",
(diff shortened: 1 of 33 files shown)


Steveb-p:
@makasim What are we going to do with the PR to `queue-interop`? Want me to take over and create a new one?

makasim:
Honestly, I don't have a good solution. do not know how to properly address it

Steveb-p:
> Honestly, I don't have a good solution. do not know how to properly address it

As discussed in the PR in queue-interop, it has been proposed to drop type for `body` parameter in interface. This would open up ability for implementations to handle message serialization on their own :smile:

IMO interface is too tight and limits possible approaches. Some transports could actually benefit from this a lot - for example Redis transport can use IGB (don't recall the proper name on top of my head) serialization that is extension-based.

Steveb-p:
and @makasim I'm only asking since @Sevavietl is currently not responding in `queue-interop` PR :smile:

PR #850, opened 2019-05-15
Add support for publisher confirms to amqp-bunny

mariusbalcytis:
RabbitMQ has publisher confirm mechanism available: https://www.rabbitmq.com/confirms.html#publisher-confirms
This guarantees that any message published to RabbitMQ will either get persisted to the disk or exception is thrown.
Without this addition, if RabbitMQ instance crashes or if you have some sort of network errors (like dropped connection), all messages published after connection is established will be lost without any notices.

Newly added mode is optional and configurable via DSN.

### DIFF
--- a/docs/transport/amqp_bunny.md
+++ b/docs/transport/amqp_bunny.md
@@ -25,6 +25,7 @@ Build on top of [bunny lib](https://github.com/jakubkulhan/bunny).
 * [Consume message](#consume-message)
 * [Subscription consumer](#subscription-consumer)
 * [Purge queue messages](#purge-queue-messages)
+* [Enable publisher confirms](#enable-publisher-confirms)
 
 ## Installation
 
@@ -265,6 +266,19 @@ $subscriptionConsumer->consume(2000); // 2 sec
 $queue = $context->createQueue('aQueue');
 
 $context->purgeQueue($queue);
+```
+
+## Enable publisher confirms
+
+If you use RabbitMQ and want to be sure your messages are received and persisted to the disk,
+you can turn on the [publisher confirms](https://www.rabbitmq.com/confirms.html#publisher-confirms).
+
+This way, after each published message, your PHP script will wait up to specified number
+of seconds to get confirmation (ACK frame) from RabbitMQ.
+
+```php
+<?php
+$factory = new AmqpConnectionFactory('amqp:?publisher_confirm_timeout=3');
 ```
 
 [back to index](../index.md)
--- a/pkg/amqp-bunny/AmqpContext.php
+++ b/pkg/amqp-bunny/AmqpContext.php
@@ -131,6 +131,9 @@ class AmqpContext implements InteropAmqpContext, DelayStrategyAware
     {
         $producer = new AmqpProducer($this->getBunnyChannel(), $this);
         $producer->setDelayStrategy($this->delayStrategy);
+        if (isset($this->config['publisher_confirm_timeout'])) {
+            $producer->enableConfirmMode((float) $this->config['publisher_confirm_timeout']);
+        }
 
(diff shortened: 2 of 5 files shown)


makasim:
cross reference https://github.com/php-enqueue/enqueue-dev/issues/749

pkg/amqp-bunny/AmqpProducer.php
```diff
@@ -50,12 +55,38 @@ class AmqpProducer implements InteropAmqpProducer, DelayStrategyAware
...
+     */
+    private $pendingDeliveryTag;
+
     public function __construct(Channel $channel, AmqpContext $context)
     {
         $this->channel = $channel;
         $this->context = $context;
     }
 
+    public function enableConfirmMode(float $timeout)
```
makasim: This method must be a part of interface, say `RabbitmqPublisherConfirmsMode`. It should define a method that enable\disable, the mode as well as a method to wait for the result. 

The interface should be implemented on context, not a producer.

The interface should be in amqp-tools for now.

docs/transport/amqp_bunny.md
```diff
@@ -267,4 +268,17 @@ $queue = $context->createQueue('aQueue');
...
+
+If you use RabbitMQ and want to be sure your messages are received and persisted to the disk,
+you can turn on the [publisher confirms](https://www.rabbitmq.com/confirms.html#publisher-confirms).
+
+This way, after each published message, your PHP script will wait up to specified number
+of seconds to get confirmation (ACK frame) from RabbitMQ.
+
+```php
+<?php
+$factory = new AmqpConnectionFactory('amqp:?publisher_confirm_timeout=3');
```
makasim: Publsher confirms is a RabbitMQ extension, works only for it. Should be prefixed `rabbitmq_`.

pkg/amqp-bunny/AmqpContext.php
```diff
@@ -131,6 +131,9 @@ public function createProducer(): Producer
     {
         $producer = new AmqpProducer($this->getBunnyChannel(), $this);
         $producer->setDelayStrategy($this->delayStrategy);
+        if (isset($this->config['publisher_confirm_timeout'])) {
+            $producer->enableConfirmMode((float) $this->config['publisher_confirm_timeout']);
```
makasim: There must be a separate option for triggering the mode globally.  Say, `rabbitmq_publisher_confirm`. There must be a default value (could be 3) for a timeout.

pkg/amqp-bunny/AmqpProducer.php
```diff
@@ -70,7 +101,7 @@ public function send(Destination $destination, Message $message): void
         InvalidMessageException::assertMessageInstanceOf($message, InteropAmqpMessage::class);
 
         try {
-            $this->doSend($destination, $message);
+            $this->doSendWithPossibleConfirm($destination, $message);
```
makasim: The feature should support two behaviors:

* Sync mode. Publish a message and wait for confirmation.
* Async mode. Publish a message return the control and let a developer check the status whenever they want.

mariusbalcytis: How should we choose the behavior? Additional method, 2 separate modes or something set in the message itself?

If 2 modes - it's probably best to make them both available globally?

Steveb-p: Probably by attaching a property / header to `Message`. I doubt `Destination` has something that you could use (but I might be wrong).

makasim: It could be a bool flag argument for `enableConfirmMode` that tells if the publishing sync or async (if async a developer should perform a check later by themselves)

PR #852, opened 2019-05-17
[SQS] Requeue with a visibility timeout

deguif:
This follows discussion started here: https://github.com/php-enqueue/enqueue-dev/commit/257fdb5fbcb95d141ced3de2fc0d1dfd2a53fb3d#r33577211


deguif:
Thanks @makasim

makasim:
Thank you too @deguif.

Issue #841, opened 2019-05-13
[Symfony] DebugClassLoader return Enqueue\Consumption\QueueConsumer::bindCallback as invalid.


vfalduto:
When execute phpUnit we have this warning 

The "Enqueue\Consumption\QueueConsumer::bindCallback()" method will require a new "string|InteropQueue $queueName" argument in the next major version of its parent class "Enqueue\Consumption\QueueConsumerInterface", not defining it is deprecated.

When we compare the class and interface, the only diff is on the phpDoc

makasim:
I cannot reproduce the issue. Feel free to send us a PR that fixes the issue.

PR #859, opened 2019-05-24
Fix bindCallback method will require new arg deprecation notice

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/841


makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/841

PR #862, opened 2019-05-24
Better Symfony doc nav

sylfabre:
Actually there is a lot of documentation for Symfony but it's not easy to navigate through the pages as the links always get you back to the general doc instead of a Symfony homepage

### DIFF
--- a/docs/bundle/async_commands.md
+++ b/docs/bundle/async_commands.md
@@ -70,4 +70,4 @@ if ($replyMessage = $promise->receive(5000)) {
 }
 ```
 
-[back to index](../index.md)
+[back to index](index.md)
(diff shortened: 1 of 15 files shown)


makasim:
Do the changes still make sense after gh-pages introduction: https://php-enqueue.github.io/ ?

Steveb-p:
> Do the changes still make sense after gh-pages introduction: https://php-enqueue.github.io/ ?

@makasim Kinda still yes. Internal bundle docs should still only go one level up instead of going to top level index.md.

Introducing index.md however is no longer necessary, since it already auto-generates table of contents now. I believe it's also why conflict is reported.

makasim:
@sylfabre could you please resolve conflicts?

Steveb-p:
@makasim Aren't they already resolved now? I used the magic "resolve conflicts" on github and went through looking what should go where... :)

makasim:
> Aren't they already resolved now? I used the magic "resolve conflicts" on github and went through looking what should go where... :)

I always forget about that feature (:

PR #863, opened 2019-05-25
Key concepts

sylfabre:
We have spent some times at AssoConnect trying to understand the logic of Enqueue for using it and creating a transport for Azure Storage Queue.

There is already a lot of content in the doc but it misses an explanation of the key concepts. It was a pain to understand how all the classes should be used: this PR tries to fix this.

It may contain incorrect info so let me know if we have misunderstood some concepts and I'll update the PR.

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -1,0 +1,69 @@
+# Key concepts
+
+If you are new to queuing system, there are some key concepts to understand to make the most of this lib.
+
+## Messages and queues
+
+The queue is where messages are sent to for later processing. Once in the queue, they wait to be fetched, processed and removed from the queue. 
+A queue may provide the following features:
+- *receive and delete delivery*: the queue deletes the message when it's fetched. If processing fails, then the message is lost and won't be processed again. This is called _at most once_ processing.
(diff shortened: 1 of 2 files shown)


Steveb-p:
Looks good. I think @makasim should confirm that this was his intended design and everything described here follows his "vision" regarding what this library should become. I'll give it another look later.

makasim:
@sylfabre @Steveb-p I've pushed some changes to the PR. Please review.

docs/concepts.md
```diff
@@ -0,0 +1,82 @@
...
+The library consist of several components. The components could be used independently or as integral part.
+
+## Components
+
+### Transport
+
+The transport is the underlying vendor-specific library that provides the queuing features: a way for programs to create, send, read messages.
+Based on [queue interop](https://github.com/queue-interop/queue-interop) interfaces. Use transport directly if you need full control or access to vendor specific features. 
+
+The most famous transports are [RabbitMQ](transport/amqp_lib.md), [Amazon SQS](transport/sqs.md), [Redis](transport/redis.md), [Filesyste](transport/filesystem.md).
```
Lctrs:
```suggestion
The most famous transports are [RabbitMQ](transport/amqp_lib.md), [Amazon SQS](transport/sqs.md), [Redis](transport/redis.md), [Filesystem](transport/filesystem.md).
```

PR #867, opened 2019-05-26
Use organization repository for doc hosting

Steveb-p:
@makasim This should allow hosting github pages under "organization" without enqueue-dev suffix (https://php-enqueue.github.io/).

I didn't do full lookup if repository split works since I have access to organization, but it should work.

### DIFF
--- a/bin/subtree-split
+++ b/bin/subtree-split
@@ -44,6 +44,7 @@ function remote()
 }
 
 remote enqueue git@github.com:php-enqueue/enqueue.git
+remote php-enqueue git@github.com:php-enqueue/php-enqueue.git
 remote simple-client git@github.com:php-enqueue/simple-client.git
 remote stomp git@github.com:php-enqueue/stomp.git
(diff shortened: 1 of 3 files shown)


bin/subtree-split
```diff
@@ -72,6 +73,7 @@ remote wamp git@github.com:php-enqueue/wamp.git
 remote monitoring git@github.com:php-enqueue/monitoring.git
 
 split 'pkg/enqueue' enqueue
+split 'docs' php-enqueue
```
makasim: could the repo be named docs ?

Steveb-p: @makasim unfortunately it's how github pages are set up - it's name has to be exactly the same as the organization.

https://help.github.com/en/articles/user-organization-and-project-pages#user-and-organization-pages-sites

makasim:
@Steveb-p fyi: the repo name should be `php-enqueue.github.io`

Steveb-p:
@makasim Sorry, I misunderstood the docs. Thanks for the info :)

PR #869, opened 2019-05-26
Github pages navigation structure

Steveb-p:
Creates basic navigation, specifically adding all transports on the list in nav. All other pages are temporarily marked as `nav_excluded` to prevent showing empty "ghost" - though working - entries.

There is a notable file added in `transports/index.md` that contains a `{:toc}` marker which can be added on pages to insert table of contents.

EDIT: Also, whitespace on line endings have been removed.


Steveb-p:
On my local build this looks like this:

image

docs/index.md
```diff
@@ -46,18 +47,18 @@ Enqueue is an MIT-licensed open source project with its ongoing development made
...
-    - [Cli commands](bundle/cli_commands.md)
-    - [Message producer](bundle/message_producer.md)
-    - [Message processor](bundle/message_processor.md)
-    - [Async events](bundle/async_events.md)
-    - [Async commands](bundle/async_commands.md)
-    - [Job queue](bundle/job_queue.md)
-    - [Consumption extension](bundle/consumption_extension.md)
-    - [Production settings](bundle/production_settings.md)
-    - [Debugging](bundle/debugging.md)
-    - [Functional testing](bundle/functional_testing.md)
```
makasim: This is a BC break. There are links to those pages on the Internet. I don't mind moving them to symfony namespace. IMO `symfony` section is a better one. but to do so, we should add pages that would guide people coming from outside to right place

Steveb-p: I've reverted this change for now. On my local build I had issues with "bundle" as directory name - it seemed that navigation markup is not read from this directory.

Steveb-p:
@makasim apparently the issue I've described in our discussion only affected my local build.

https://php-enqueue.github.io/symfony

Thanks for merging :)

PR #881, opened 2019-06-01
Add composer runnable scripts for PHPStan & PHP-CS

Steveb-p:
This allows for forks in composer requiring whole package (enqueue) to
work properly.

Also add code style commands as composer scripts for ease of use.

This allows calling for example `composer run cs-fix some-file.php`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,6 +3,11 @@
   "type": "project",
   "minimum-stability": "beta",
   "homepage": "https://enqueue.forma-pro.com/",
+  "scripts": {
+    "cs-fix": "bin/php-cs-fixer fix --config=.php_cs.php",
+    "cs-lint": "bin/php-cs-fixer fix --config=.php_cs.php --no-interaction --dry-run --diff",
+    "phpstan": "bin/phpstan analyse --memory-limit=512M -c phpstan.neon"
+  },
   "require": {
     "php": "^7.1.3",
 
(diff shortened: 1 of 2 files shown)


makasim:
Not sure I get the purpose of this PR, @Steveb-p could you elaborate on why we might need it?

Steveb-p:
@makasim The name for package in composer.json should match the name used in packagist to allow proper forking / repository override.

Use case for this is using your owned version of a library instead of the main one available in packagist - which is helpful when applying some small, custom patches without intent of maintaining a separate package.

I do realize that there should really be no case that whole `enqueue-dev` repository is used (just sub-repositories instead), but I do see that there are downloads done via packagist.

https://packagist.org/packages/enqueue/enqueue/stats

makasim:
tps://packagist.org/packages/enqueue/enqueue/stats is a valid package located here https://github.com/php-enqueue/enqueue-dev/tree/master/pkg/enqueue

The enqueue-dev repo is not even registred on packagist.

Steveb-p:
Then the PR can be discarded. 

I can push those scripts declared in composer as a separate PR if you like or drop them entirely - or rename this PR - though I believe they might become helpful for potential contributors - allowing new ones to easily call common tasks without knowing their exact configuration / reading them from travis configuration file. I can also add a note to contributing.md regarding their use.

makasim:
> I can push those scripts declared in composer as a separate PR if you like or drop them entirely - or rename this PR 

That would be great. I am fine with either way.

Steveb-p:
@makasim I've changed this PR into a one that introduces composer scripts instead only. Figured it would be ok if above discussion is kept instead of creating new PR.

Issue #884, opened 2019-06-06
Documentation - quick tour


Kon-Xen:
Hi,
Noticed that in the quick tour of the documentation for the symfony bundle
The example instructions suggest for the config 
```
enqueue:
    default:
        transport: "amqp:"
        client: ~
```
that gives a yaml syntax error....
Could it be that instead the example should read ? 
```
enqueue:
    transport:
        default: "amqp:"
    client: ~
```
Best,

K.

Steveb-p:
@Kon-Xen No, the second example would be wrong for the current version. It was correct for 0.8.x. Since then clients and transports are tied under a single key.

Kon-Xen:
what would be the correct syntax then ?

Steveb-p:
> what would be the correct syntax then ?

Check for some common yaml errors in your configuration, like mixing tabs & spaces (yaml requires you to use one or the other for indentation, but not both).

If it were a misconfiguration issue, you would have a different exception, for example:  
`Fatal error: Uncaught Symfony\Component\Config\Definition\Exception\InvalidConfigurationException: Unrecognized option "dsn" under "enqueue.transport"`

Also check if you're not using obsolete 0.8.x version for enqueue bundle, since the configuration presented in quick tour is for current 0.9.x.

Kon-Xen:
Thank you for your prompt responses and 
All understood. (and I have checked spaces & tabs and what not & Managed to figure things out after many hours)
There is a confusion however. 
a. Yes it turns out I am using 8.42 ver.
b. see The config example here: https://github.com/php-enqueue/enqueue-dev/blob/master/docs/bundle/quick_tour.md
and compare to here: https://php-enqueue.github.io/bundle/config_reference/

Steveb-p:
@Kon-Xen both https://php-enqueue.github.io/bundle/quick_tour/ and https://php-enqueue.github.io/bundle/config_reference/ look fine to me (well, except that we're already working on expanding comments in config reference to better explain what options do what).

What do you mean?

Kon-Xen:
If you say they are fine,  they are fine. 

Now, if you fancy a chat I can try to elaborate...
Assuming that I understood correctly,
https://github.com/php-enqueue/enqueue-dev/blob/master/docs/bundle/quick_tour.md
is confusing me because:
a) wording,  "First, you have to configure a transport layer and **set one** to be default" 
     Do I have to set up more than one transports?
[...]

PR #885, opened 2019-06-06
[Bundle] Expand quick tour for Symfony Bundle

Steveb-p:
According to comments in #884.


makasim:
This is a really nice one! Thank you @Steveb-p

PR #888, opened 2019-06-07
Fix PHPStan error from Symfony 4.3 Event Dispatcher

Steveb-p:
Also moves PHPStan level declaration from command line to configuration file.

Note that `EventDispatcher`'s constructor is basically a no-op, unless instantiated class is actually an instance of `EventDispatcher` itself, so we might be better off silencing this PhpStan error instead?

### DIFF
--- a/bin/pre-commit
+++ b/bin/pre-commit
@@ -131,7 +131,7 @@ function runPhpstan()
     $returnCode = null;
 
     exec(sprintf(
-        'docker run --workdir="/mqdev" -v "`pwd`:/mqdev" --rm enqueue/dev:latest php -d memory_limit=1024M bin/phpstan analyse -l 1 -c phpstan.neon %s',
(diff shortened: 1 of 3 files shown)


makasim:
If it does not have a constructor, can we remove parent::__construct call?

Steveb-p:
> If it does not have a constructor, can we remove parent::__construct call?

Symfony 4.3 introduces a constructor, that "does nothing" when the class being instantiated is not exactly `EventDispatcher`. Previously there was no constructor at all.

Symfony 4.2: no `__construct`
https://github.com/symfony/event-dispatcher/blob/4.2/EventDispatcher.php

Symfony 4.3:
https://github.com/symfony/event-dispatcher/blob/4e6c670af81c4fb0b6c08b035530a9915d0b691f/EventDispatcher.php#L39-L44

This change is mainly targeted at conforming to PHPStan, and calling parent constructor if it's present.

makasim:
@Steveb-p now the change makes sense to me. Thank you for clarifications.

Steveb-p:
Current checks in travis are done against Symfony 3.4 and subsequently fail.

Guess a different approach will be needed :)

PR #901, opened 2019-06-14
deserialize sqs message attributes

bendavies:
Allow deserializing all messages attributes, which we can use for logging and debugging.

### DIFF
--- a/pkg/sqs/SqsMessage.php
+++ b/pkg/sqs/SqsMessage.php
@@ -24,6 +24,11 @@ class SqsMessage implements Message
     private $headers;
 
     /**
+     * @var array
+     */
+    private $attributes;
+
+    /**
      * @var bool
      */
     private $redelivered;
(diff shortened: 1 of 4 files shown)


pkg/sqs/SqsMessage.php
```diff
@@ -113,6 +119,21 @@ public function getHeader(string $name, $default = null)
...
+    }
+
+    public function getAttributes(): array
+    {
+        return $this->attributes;
+    }
+
+    public function getAttribute(string $name, $default = null)
+    {
+        return array_key_exists($name, $this->attributes) ? $this->attributes[$name] : $default;
```
Steveb-p: Wouldn't `isset` be sufficient? It is said it's a lot faster than `array_key_exists` with the difference that it will report `null` values as not set.

https://ilia.ws/archives/247-Performance-Analysis-of-isset-vs-array_key_exists.html
http://thinkofdev.com/php-fast-way-to-determine-a-key-elements-existance-in-an-array/

At least it was in 2012, but I don't think this changed.

bendavies: sure there are better ways now, `return $this->attributes[$name] ?? $default`, but i was copying the current style of the code base.

PR #905, opened 2019-06-20
Add lazy queue support for RabbitMQDriver

### LINKED ISSUE #904: Supporting lazy queues with the rabbit MQ driver
Hi there.

I'm working with Enqueue, and using RabbitMQ as the underlying queue for messages, and I'm looking for some pointers on supporting [Lazy Queues](https://www.rabbitmq.com/lazy-queues.html) with Enqueue.

### What Lazy queues are and what they are useful.

The best practice recommendations for RabbitMQ that I see, recommend you keep queues short, but there are some cases where you might want support long queues.

> Since RabbitMQ 3.6.0, the broker has the concept of Lazy Queues - queues that move their contents to disk as early as practically possible, and only load them in RAM when requested by consumers, therefore the lazy denomination.
> One of the main goals of lazy queues is to be able to support very long queues (many millions of messages).

I have a use case where I need to put a lot work onto a queue (i.e. a few million messages, for workers to run through), and if you're using rabbit 3.6 or later, you can do this by passing in  arguments like this `$queue->setArguments(['x-queue-mode' => 'lazy']);` when creating a  queue.

### Creating a lazy RabbitMQ driver

I know we have a [RabbitMQ driver](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Client/Driver/RabbitMqDriver.php), activated by using a connection string like `amqp+rabbitmq` as our DSN.

```
class RabbitMqDriver extends AmqpDriver
{
    /**
     * @return AmqpQueue
     */
    protected function doCreateQueue(string $transportQueueName): InteropQueue
    {
        $queue = parent::doCreateQueue($transportQueueName);
        $queue->setArguments(['x-max-priority' => 4]);
        return $queue;
    }
}

```

I'm proposing something like passing in a connection string like `amqp+rabbitmqlazy` for our DSN or similar, and a _lazy_ driver like so:

```
class LazyRabbitMqDriver extends AmqpDriver
{
    /**
     * @return AmqpQueue
     */
    protected function doCreateQueue(string $transportQueueName): InteropQueue
    {
        $queue = parent::doCreateQueue($transportQueueName);
        $queue->setArguments(['x-max-priority' => 4]);
        $queue->setArguments(['x-queue-mode' => 'lazy']);
        return $queue;
    }
}
```

### Specific Questions

 - 1) Does lazy refer to something else here? I see [`lazy` referred to in the config reference in the docs](https://github.com/php-enqueue/enqueue-dev/blob/742a0af0cb6986e68f7109647a9effcdbc16a582/docs/bundle/config_reference.md
), but I see no mention in the code base of `x-queue-mode`,  nor as [a supported feature](
https://github.com/php-enqueue/enqueue-dev/blob/742a0af0cb6986e68f7109647a9effcdbc16a582/docs/client/supported_brokers.md)
 so I suspect it's something different.

- 2) Where do I set up the extension and the driver to load in? Would I add it to [Resources.php in enqueue/Client/](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Client/Resources.php#L60-L65)

mrchrisadams:
This PR should resolve #904, but I think I need some pointers about setting up the test properly, and code style.

How should I be passing in a config option in this case - would I pass in arbitrary values to `createDummyConfig` or something similar or would I create a separate DummyConfig instead?

### DIFF
--- a/pkg/enqueue/Client/Driver/RabbitMqDriver.php
+++ b/pkg/enqueue/Client/Driver/RabbitMqDriver.php
@@ -16,16 +16,14 @@ class RabbitMqDriver extends AmqpDriver
         $queue = parent::doCreateQueue($transportQueueName);
 
         $driverOptions = $config->getDriverOptions();
-        $islazyQueue = false;
-        $lazyQueueArray = array_filter($driverOptions, function ($arr) {
-            return array_key_exists('rabbit_mq_lazy_queues', $arr);
-        });
+        $lazyQueueArray = $driverOptions['rabbit_mq_lazy_queues'] ?? [];
+        $isLazyQueue = in_array($transportQueueName, $lazyQueueArray, true);
 
-        $islazyQueue = array_key_exists($transportQueueName, $lazyQueueArray[0]['rabbit_mq_lazy_queues']);
-
-        $queue->setArguments(['x-max-priority' => 4]);
-        if ($islazyQueue) {
-            $queue->setArguments(['x-queue-mode' => 'lazy']);
+        if (true == $isLazyQueue) {
+            $queue->setArguments(
+                ['x-queue-mode' => 'lazy', 'x-max-priority' => 4]);
+        } else {
+            $queue->setArguments(['x-max-priority' => 4]);
         }
 
         return $queue;
--- a/pkg/enqueue/Client/Driver/RabbitMqDriver.php
+++ b/pkg/enqueue/Client/Driver/RabbitMqDriver.php
@@ -15,11 +15,16 @@ class RabbitMqDriver extends AmqpDriver
         $config = parent::getConfig();
         $queue = parent::doCreateQueue($transportQueueName);
 
-        $queue->setArguments(['x-max-priority' => 4]);
+        $driverOptions = $config->getDriverOptions();
+        $islazyQueue = false;
+        $lazyQueueArray = array_filter($driverOptions, function ($arr) {
+            return array_key_exists('rabbit_mq_lazy_queues', $arr);
+        });
 
-        // we use isLazy to avoid collisions with 'lazy' which
-        // is already used to meansomething else
-        if (true == $config->getTransportOption('isLazy')) {
+        $islazyQueue = array_key_exists($transportQueueName, $lazyQueueArray[0]['rabbit_mq_lazy_queues']);
+
+        $queue->setArguments(['x-max-priority' => 4]);
+        if ($islazyQueue) {
             $queue->setArguments(['x-queue-mode' => 'lazy']);
         }
 
--- a/pkg/enqueue/Client/Driver/RabbitMqDriver.php
+++ b/pkg/enqueue/Client/Driver/RabbitMqDriver.php
@@ -19,11 +19,10 @@ class RabbitMqDriver extends AmqpDriver
         $lazyQueueArray = $driverOptions['rabbit_mq_lazy_queues'] ?? [];
         $isLazyQueue = in_array($transportQueueName, $lazyQueueArray, true);
 
-        if (true == $isLazyQueue) {
-            $queue->setArguments(
-                ['x-queue-mode' => 'lazy', 'x-max-priority' => 4]);
-        } else {
-            $queue->setArguments(['x-max-priority' => 4]);
+        $queue->setArguments(['x-max-priority' => 4]);
+
+        if ($isLazyQueue) {
+            $queue->setArgument('x-queue-mode', 'lazy');
         }
 
         return $queue;
--- a/pkg/enqueue/Client/Driver/RabbitMqDriver.php
+++ b/pkg/enqueue/Client/Driver/RabbitMqDriver.php
@@ -12,8 +12,16 @@ class RabbitMqDriver extends AmqpDriver
      */
     protected function doCreateQueue(string $transportQueueName): InteropQueue
     {
+        $config = parent::getConfig();
         $queue = parent::doCreateQueue($transportQueueName);
+
         $queue->setArguments(['x-max-priority' => 4]);
+
+        // we use isLazy to avoid collisions with 'lazy' which
+        // is already used to meansomething else
+        if($config->getTransportOption('isLazy') == true) {
+            $queue->setArguments(['x-queue-mode' => 'lazy']);
+        }
 
         return $queue;
     }
--- a/pkg/enqueue/Client/Driver/RabbitMqDriver.php
+++ b/pkg/enqueue/Client/Driver/RabbitMqDriver.php
@@ -19,7 +19,7 @@ class RabbitMqDriver extends AmqpDriver
 
         // we use isLazy to avoid collisions with 'lazy' which
         // is already used to meansomething else
-        if($config->getTransportOption('isLazy') == true) {
+        if (true == $config->getTransportOption('isLazy')) {
             $queue->setArguments(['x-queue-mode' => 'lazy']);
         }
 
--- a/pkg/enqueue/Tests/Client/Driver/RabbitMqDriverTest.php
+++ b/pkg/enqueue/Tests/Client/Driver/RabbitMqDriverTest.php
@@ -68,25 +68,35 @@ class RabbitMqDriverTest extends TestCase
         $context
             ->expects($this->once())
             ->method('createQueue')
-            ->willReturn($this->createQueue('aName'))
+            ->willReturn($this->createQueue('aname'))
         ;
+
+        $config = Config::create(
+            null,
+            null,
+            null,
+            null,
+            null,
+            null,
+            null,
+            [
+                'dsn' => getenv('RABBITMQ_AMQP_DSN'),
+            ],
+            ['rabbit_mq_lazy_queues' => ['aname']]
+        );
 
         $driver = $this->createDriver(
             $context,
             // presumably we pass in some isLazy check here
-            $this->createDummyConfig(),
+            $config,
             new RouteCollection([])
         );
 
         /** @var AmqpQueue $queue */
-        $queue = $driver->createQueue('aName');
-
-        $this->assertSame([
-            'x-max-priority' => 4,
-            'x-queue-mode' => 'lazy',
-            ],
-            $queue->getArguments()
-        );
+        $queue = $driver->createQueue('aname');
+        $queueArguments = $queue->getArguments();
+        $this->assertSame($queueArguments['x-queue-mode'], 'lazy');
+        $this->assertSame($queueArguments['x-max-priority'], 4);
     }
 
     protected function createDriver(...$args): DriverInterface
--- a/pkg/enqueue/Tests/Client/Driver/RabbitMqDriverTest.php
+++ b/pkg/enqueue/Tests/Client/Driver/RabbitMqDriverTest.php
@@ -66,10 +66,12 @@ class RabbitMqDriverTest extends TestCase
     {
         $context = $this->createContextMock();
         $context
-            ->expects($this->once())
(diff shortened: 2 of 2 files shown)


makasim:
There is a better way based on [driver_options](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Symfony/Client/DependencyInjection/ClientFactory.php#L77)

mrchrisadams:
Thanks @makasim  - I'm afraid I didn't understand your comment - are you saying I should put these into the `driver_options` array in the next argument for `Config::create`, so we have something like this instead?

```
$config = Config::create(
            null,
            null,
            null,
            null,
            null,
            null,
            null,
            [
                'dsn' => getenv('RABBITMQ_AMQP_DSN'),
            ],
            ['isLazy' => true]
        );
```

Are there any examples in the test suite that i can refer to, to see how I should be doing this?

Thanks

makasim:
There are no examples. You are the first who goes this way. So only the source code. Yeap, you need the last argument of the config object. 

1. Pass the option `rabbitmq_[queue_name]_x_queue_mode`. 
2. Adjust `doCreateQueue` method to check if this option set, and if so, pass it as argument to queue.

mrchrisadams:
Hey @makasim I've got some code together that's working locally, and I've a test for it now, but I'm having a hard time getting the tests running locally.  Here's some sample output - have you seen a case of thruway failing like this before and stopping the tests?

```
time ./bin/test.sh --filter "testShouldCreateLazyQueueifisLazySetinConfig"
+ set -e
+ docker-compose run --workdir=/mqdev --rm dev ./docker/bin/test.sh --filter testShouldCreateLazyQueueifisLazySetinConfig
Starting enqueue-dev_mongo_1 ... done
Starting enqueue-dev_beanstalkd_1  ... done
Starting enqueue-dev_mysql_1       ... done
Starting enqueue-dev_gearmand_1    ... done
Starting enqueue-dev_thruway_1       ... done
Starting enqueue-dev_zookeeper_1     ... done
Starting enqueue-dev_rabbitmq_1      ... done
Starting enqueue-dev_rabbitmq_1    ...
Starting enqueue-dev_kafka_1         ... done
Starting enqueue-dev_google-pubsub_1 ... done
Starting enqueue-dev_redis_1         ... done
service is online rabbitmq:5672
service is online rabbitmqssl:5671
service is online mysql:3306
service is online postgres:5432
service is online redis:6379
service is online beanstalkd:11300
service is online gearmand:4730
service is online kafka:9092
service is online mongo:27017
thruway: forward host lookup failed: Unknown host
wait for service thruway:9090
thruway: forward host lookup failed: Unknown host
wait for service thruway:9090
thruway: forward host lookup failed: Unknown host
wait for service thruway:9090
thruway: forward host lookup failed: Unknown host
wait for service thruway:9090
^Cthruway: forward host lookup failed: Unknown host
wait for service thruway:9090
./bin/test.sh --filter "testShouldCreateLazyQueueifisLazySetinConfig"  0.72s user 0.23s system 12% cpu 7.715 total
```

makasim:
try to comment it out temporarily https://github.com/php-enqueue/enqueue-dev/blob/master/docker/bin/test.sh#L41

pkg/enqueue/Client/Driver/RabbitMqDriver.php
```diff
@@ -12,8 +12,21 @@ class RabbitMqDriver extends AmqpDriver
...
+        $config = parent::getConfig();
         $queue = parent::doCreateQueue($transportQueueName);
+
+        $driverOptions = $config->getDriverOptions();
+        $islazyQueue = false;
+        $lazyQueueArray = array_filter($driverOptions, function ($arr) {
+            return array_key_exists('rabbit_mq_lazy_queues', $arr);
+        });
+
+        $islazyQueue = array_key_exists($transportQueueName, $lazyQueueArray[0]['rabbit_mq_lazy_queues']);
```
Steveb-p:
```php
$lazyQueueArray = $driverOptions['rabbit_mq_lazy_queues'] ?? [];
$isLazyQueue = $lazyQueueArray[$transportQueueName] ?? false;
```
A lot nicer and less error prone. Original code with `array_filter` can potentially break if `rabbit_mq_lazy_queues` options is not on first position and generally works slower due to the fact that `array_key_exists` is known to be roughly 5x slower that `isset`.

pkg/enqueue/Client/Driver/RabbitMqDriver.php
```diff
@@ -16,16 +16,14 @@ protected function doCreateQueue(string $transportQueueName): InteropQueue
...
+        $isLazyQueue = in_array($transportQueueName, $lazyQueueArray, true);
 
-        $islazyQueue = array_key_exists($transportQueueName, $lazyQueueArray[0]['rabbit_mq_lazy_queues']);
-
-        $queue->setArguments(['x-max-priority' => 4]);
-        if ($islazyQueue) {
-            $queue->setArguments(['x-queue-mode' => 'lazy']);
+        if (true == $isLazyQueue) {
+            $queue->setArguments(
+                ['x-queue-mode' => 'lazy', 'x-max-priority' => 4]);
```
Steveb-p: Use `$queue->setArgument(key, value)` instead. That way you can drop the inconvenient `if...else`.

```php
$queue->setArguments(['x-max-priority' => 4]);
if ($isLazyQueue) {
        $queue->setArgument('x-queue-mode' => 'lazy');
}
```
This ensures that adding new options is kept simple and does not influence too many lines in version control (not to mention creating a devilish `if` :smiley: )

`true == $isLazyQueue` is not really needed. PHP will cast `$isLazyQueue` to boolean value anyway, and since there is no type check there is no point to write `true ==`.

Lastly, even if you'd like to stick to `setArguments` method, options array should be spread across lines. Like so:
```php
$queue->setArguments([
    'x-queue-mode' => 'lazy', 
    'x-max-priority' => 4,
]);
```
Reason being, if anyone decides to add another option, it will appear as a new line and will not cause conflicts between different PRs. This is also the reason that last array element has a comma at the end.

pkg/enqueue/Tests/Client/Driver/RabbitMqDriverTest.php
```diff
@@ -62,6 +62,43 @@ public function testShouldCreateQueueWithMaxPriorityArgument()
...
+            // presumably we pass in some isLazy check here
+            $config,
+            new RouteCollection([])
+        );
+
+        /** @var AmqpQueue $queue */
+        $queue = $driver->createQueue('aname');
+        $queueArguments = $queue->getArguments();
+        $this->assertSame($queueArguments['x-queue-mode'], 'lazy');
+        $this->assertSame($queueArguments['x-max-priority'], 4);
```
Steveb-p: Add a check for a `Queue` that has a different name, since we're here, proving that `x-queue-mode` will not appear in queues not listed in `$config` as lazy.

```php
$queue = $driver->createQueue('bname');
$queueArguments = $queue->getArguments();
$this->assertFalse(isset($queueArguments['x-queue-mode']));
$this->assertSame($queueArguments['x-max-priority'], 4);
```

Remove the comment `// presumably we pass in some isLazy check here` just to keep it clean and in my opinion you're done :smile: 

We might add an info about this functionality in the docs, if you or someone else finds a suitable place :smile:

pkg/enqueue/Client/Driver/RabbitMqDriver.php
mrchrisadams: Cool, thanks for the pointers @Steveb-p  - PHP is definitely not my primary language, so these are really helpful.

pkg/enqueue/Tests/Client/Driver/RabbitMqDriverTest.php
mrchrisadams: I'm happy to add this but I'm not sure where it makes sense to add this in the docs. Would you accept this as a separate PR for the docs instead?

Steveb-p: @mrchrisadams Don't duplicate the test :smile: Just add the four lines I've posted above at the end of this test :smile:

Steveb-p: @mrchrisadams 

> I'm happy to add this but I'm not sure where it makes sense to add this in the docs. Would you accept this as a separate PR for the docs instead?

Yes, I'll follow this PR with opening an issue, so we can take our time adding proper documentation. I believe it's a functionality that you need, so docs "can wait".

Moreover, if makasim will want to wait before publishing new version - for whatever reason, even if simply to not publish for each and every small functionality - I can show you how to override repositories so you can use your enqueue build immediately, before a new version is officially out.

See https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository

mrchrisadams:
Hi folks,

Until this PR is merged into a new release, I'm not sure I understand how to use this code in projects, and I think I might need some pointers - hope that's okay.

I've read the instructions about [Loading a package from a VCS repository](https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository), here, and I've created a branch accordingly with the correct name, and updated my composer.json in a branch of an app that has so far been using the enqueue-bundle library (I assume enqueue-bundle uses some code from enqueue-bundle somewhere, but it wasn't clear how it consumed them).

You can see the composer.json file here:

https://github.com/thegreenwebfoundation/thegreenwebfoundation/blob/ca-separate-queue/apps/api/composer.json#L41

I think I'm doing something wrong though, as when running `composer update` it looks like the enqueue-bundle requires other extensions that I'm not using in production:

```
composer update
Restricting packages listed in "symfony/symfony" to "4.2.*"
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for enqueue/enqueue-dev dev-ca-rabbitmqdriver-lazy-support -> satisfiable by enqueue/enqueue-dev[dev-ca-rabbitmqdriver-lazy-support].
    - enqueue/enqueue-dev dev-ca-rabbitmqdriver-lazy-support requires ext-gearman ^2.0 -> the requested PHP extension gearman is missing from your system.

  To enable extensions, verify that they are enabled in your .ini files:
    - /usr/local/etc/php/7.3/php.ini
    - /usr/local/etc/php/7.3/conf.d/ext-opcache.ini
  You can also run `php --ini` inside terminal to see which files are used by PHP in CLI mode.
```

If I want to use this for implementing lazy queues, how do I require just the bits I need (i.e. the RabbitMQ bits, but not gearman?).

Thanks

Steveb-p:
@mrchrisadams you will need to fork https://github.com/php-enqueue/enqueue specifically and apply your changes.

Reason why additional extensions are required that you see is that `enqueue-dev` is a so-called mono-repository. It contains all code relevant to `enqueue` project (all possible drivers, a bundle, magento & laravel integration and so on). This is done so any changes made to each and every part of enqueue is compatible and relevant at any given time.

You can read more about monorepositories by googling it. There is also a nice https://github.com/korfuri/awesome-monorepo that you can visit which contains an opinionated list of anything related to monorepositories, be it tools or articles.

You can see the code responsible for spliting to subrepositories in this repository here: https://github.com/php-enqueue/enqueue-dev/blob/master/bin/subtree-split if you're curious.

Your composer definition file is perfect. All it need is just a fork of subrepository, not the monorepo.

```json
{
     "type": "vcs",
     "url": "https://github.com/thegreenwebfoundation/enqueue-dev"
}
```
Replace https://github.com/thegreenwebfoundation/enqueue-dev with a fork of abovementioned repository.

If you have any questions, I'm available on gitter.

PR #906, opened 2019-06-21
Adding generic driver for schema SNS

Nyholm:
I could not really configure SNS without this.


Nyholm:
Thank you for merging

PR #914, opened 2019-07-11
async_commands: extended configuration proposal

uro:
Hi guys, I use async_commands and I found out that it's missing a few things. 

Firstly, we can't prefix `run_command` queue and because of that, on example: if we use one queue provider with many services it will provide problems as we will push everything to one queue. 
Same with many configurations, each of it can have a different prefix. 

Then, the second issue: default Symfony process timeout is 60 seconds. after this, the message is requeued and processed again. We don't have any option to extend this time. 

**In this pull-request I provide a proposal for these two issues:**
- Prefix `run_command` queue
- Change the default process timeout


uro:
Hey @Steveb-p. Did you have any chance to look at it? The build is failing from 3rd party reason. I'd like to know if we can do some progress with it

Steveb-p:
@uro I've re-run the failing test and it is indeed steming from a service failing to start. Green now.

I'd like @makasim to take a quick look on it as well, just to be sure, and I think it's good to go :+1:

pkg/async-command/DependencyInjection/AsyncCommandExtension.php
```diff
@@ -12,13 +12,22 @@ class AsyncCommandExtension extends Extension
...
+            $id = sprintf('enqueue.async_command.%s.run_command_processor', $client['name']);
             $container->register($id, RunCommandProcessor::class)
-                ->addArgument('%kernel.project_dir%')
+                ->addArgument('%kernel.project_dir%', $client['timeout'])
                 ->addTag('enqueue.processor', [
-                    'client' => $client,
-                    'command' => Commands::RUN_COMMAND,
-                    'queue' => Commands::RUN_COMMAND,
+                    'client' => $client['name'],
+                    'command' => $client['prefix'].Commands::RUN_COMMAND,
```
makasim: IMO prefix is misleading, what about to allow configure a whole command and queue names?

makasim: The default value could be the one from `Commands::RUN_COMMAND`

uro: @makasim  sounds like a great idea. I'm going to implement the fix

uro:
Hey @makasim, please accept my apologies for this late reply. I submitted the change related to the command name. 

Could you take a look again?

Steveb-p:
Looks good to me. There should be no BC breaks because of it so no migration documentation necessary - though docs related to async commands might need an update.

PR #916, opened 2019-07-16
DBAL Change ParameterType class to Type class

Nevoss:
Doctrine DBAL version 2.6 does not have ParameterType class.
and the library supports version 2.6 of Doctrine DBAL.

Hope it will be merge soon.
Thanks.


Steveb-p:
We have to maintain BC to previous versions, at least until they are deprecated / stop being commonly used. So we will probably have to either check against class existence via `class_exists` or check doctrine package version.

Issue #919, opened 2019-07-18
Multiple Queues in Symfony


lostfocus:
There are a couple of open _and_ closed issues regarding the documentation when it comes to multiple queues with the Symfony bundle. I'm absolutely willing to write a nice .md-file to document how it works if someone could just point me in the right direction.

Let's say I have two topics I want to publish to: `hello` and `goodbye` - so far that's easy:

`Enqueue\Client\ProducerInterface $producer` in the dependency injection part of my service and then:

```
$helloWorld = new Enqueue\Client\Message(['object' => 'world']);
$producer->sendEvent('hello',$helloWorld);
```

and

```
$goodbyeWorld = new Enqueue\Client\Message(['object' => 'world']);
$producer->sendEvent('goodbye',$goodbyeWorld);
```

Now when I look into my queue, I can see that both messages are in the queue `enqueue.app.default` and when I run the queue runner with `bin/console enqueue:consume --time-limit="now+5minutes" -vvv` I can see that the processors find the messages and do their thing.

Now what I'd like to do is to execute `bin/console enqueue:consume --time-limit="now+5minutes" -vvv hello` and `bin/console enqueue:consume --time-limit="now+5minutes" -vvv goodbye` separately - and I figure that works only when they're not both in the `enqueue.app.default` queue. 

_At what point can I tell the message or the producer which queue to use?_

As I said - just point me in the right direction and I'm happy to write a nice explanation for the documentation.

Steveb-p:
@lostfocus https://github.com/php-enqueue/enqueue-dev/tree/master/docs/bundle should be a good place for it. It will be converted into https://github.com/php-enqueue/enqueue-dev/tree/master/docs/bundle once it's merged.

Also see https://php-enqueue.github.io#contributing-to-this-documentation if you want to see how it will look locally. You can use local Jekyll of course if you don't like docker with shared directories.

makasim:
You can configure a queue on processor registration. It could be an attribute of service tag, a parameter of queue subscriber, a parameter of command subscriber. All end up in https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Client/Route.php.

Some examples could be found here: https://github.com/php-enqueue/enqueue-dev/blob/master/docs/bundle/message_processor.md

lostfocus:
Thanks. That will be my weekend project then.

Issue #838, opened 2019-05-09
Too minimalistic approach for config documentation


damianprzygodzki:
I am totally confused with documentation that is constructed really poor.

I understand that we have topics to connect messages to processors, but there is no information about connecting this to transport queue. There is also no info about ability of configuration more clients (Clients or topics?). Naming in config is also confusing, like 'routerQueue' vs. 'defaultQueue'.

Wasted lot of time to research it on my own...

Steveb-p:
You're welcome to provide improvements to documentation.

As with many open-source projects, oftentimes documentation is satisfactory to the one writing the actual library, but not for end users - which is obviously due to the fact that author does have an understanding of what code does and "assumes" knowledge of it's internals.

At the very least please provide links to documentation pages that were confusing for you.

damianprzygodzki:
> You're welcome to provide improvements to documentation.

It is totally hopeless idea to let person who struggle with bundle create documentation.

> As with many open-source projects, oftentimes documentation is satisfactory to the one writing the actual library, but not for end users

That's why i created issue - to announce that there is a need.

> At the very least please provide links to documentation pages that were confusing for you.

It is also funny because each of documentation available (official github repo or readthedocs service) has different and not up to date docs, so if you need URL, each one available is confusing.

Issue is about config documentation, config is documented in one chapter, how can i be more precise?

https://github.com/php-enqueue/enqueue-dev/blob/master/docs/bundle/config_reference.md

Steveb-p:
> It is totally hopeless idea to let person who struggle with bundle create documentation.

Quite the contrary. You were looking for a particular option or configuration value and was unable to deduce what it stands for or where it can be found.  
If you have it figured out - you said you spent time on it - then you're best suited to provide contribution since your knowledge is "fresh". If you did not, then ask a specific question.

As I said - even though I'm only "responsible" for Kafka part of library - I can imagine it might be difficult for maintainer to cater to everyone. Usually https://github.com/php-enqueue/enqueue-dev/blob/master/docs/bundle/quick_tour.md file is sufficient for most basic configurations, configuration reference that you have linked is just that - a config reference. Certainly it can be improved, but keep in mind we're spending our private time doing so - so pointing us in the right direction as specifically as possible is more than welcome.

[...]

Issue #806, opened 2019-03-26
Docs issue for multiple queues in Symfony bundle


Toflar:
Hey folks,

I'm trying to configure the bundle in such a way that I can get 3 queues on redis (to mimic priority support).
So what I would like to have is the following

* 3 producers that all produce to the same transport but a different queue
* 3 workers `bin/console enqueue:consume` that consume from the same transport but a different queue

How do I configure that? I don't get it from the docs. I'll happily send a PR to fix that once I get it 😄

makasim:
There is [client-queue-names](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Symfony/Client/ConsumeCommand.php#L86) argument for that. Please note that it is a client queue name, not a transport one. You can find out which one to use by running `enqueue:routes` command

makasim:
There is similar options for `enqueue:transport:consume` command.

Toflar:
I saw that but by just reading the docs I don‘t know how to configure multiple queues for the same transport. The docs is basically just a `config:dump-reference` which doesn‘t say a lot about usage. It‘s lacking examples!

makasim:
I will add some when got into that topic again. feel free to contribute by improving the doc too. You are the one who understands the best what is not clear and hard to grasp.

Issue #835, opened 2019-05-01
Message processor without default client


koderoff:
I setup enqueue for two clients something like this:

/config/packages/enqueue.yaml

```
enqueue:
    amqp:
        transport: '%env(resolve:ENQUEUE_DSN_AMQP)%'
        client: ~            
    kafka:
        transport: '%env(resolve:ENQUEUE_DSN_KAFKA)%'
        client: ~
```

/config/services.yaml

```
App\Async\ClickMessageProcessor:
    tags:
        - { name: 'enqueue.command_subscriber', client: 'kafka' }
```
Now I have two message processors which implements CommandSubscriberInterface and I want to bind each of them to his specific client. But the problem is that each processor that implements CommandSubscriberInterface is automatically registered with amqp client (which is default in my case). So, as a result I have ClickMessageProcessor that is registered with amqp client as well as kafka.

Is there a way to deal with this?

I am aware of **enqueue.processor** tag but it is not suits me because of losing flexibility of setting up the command

Steveb-p:
You can disable `autoconfig` in `ClickMessageProcessor` service, which should turn off assigning tags based on interfaces they implement.

koderoff:
Exactly what is needed. Thanks)

Steveb-p:
@koderoff for your reference, you can always check what configuration Symfony actually applied to a service using `php bin/console debug:container` command, even so far as to specify a particular service you want to look at (`php bin/console debug:container App\\Async\\ClickMessageProcessor --tags` I believe?)

There you would notice additional tag being assigned :)

PR #920, opened 2019-07-20
documentation for multiple queues with the symfony bundle

lostfocus:
As mentioned in #919 - and in #806 and #838 

There is not really any new information on this page but it should answer this pretty common question which I have found unanswered previously both here and on StackOverflow.

### DIFF
--- a/docs/bundle/multiple_queues.md
+++ b/docs/bundle/multiple_queues.md
@@ -1,7 +1,7 @@
 ---
 layout: default
 parent: "Symfony bundle"
-title: Functional testing
+title: Multiple queues
 nav_order: 13
 ---
 {% include support.md %}
--- a/docs/bundle/multiple_queues.md
+++ b/docs/bundle/multiple_queues.md
@@ -1,0 +1,70 @@
+---
+layout: default
+parent: "Symfony bundle"
+title: Functional testing
+nav_order: 13
+---
+{% include support.md %}
+
+# Multiple queues
+
+One of the most powerful features of this bundle is the ability to run multiple queues with the same transport/client. 
+
+## Different queues for topic routing and processing
+
+In the default configuration both the topic routing and the job processing tasks are written to the `default` queue. If you want to separate those two types of tasks you'll have to change the [configuration](config_reference.md) in `config/packages/enqueue.yaml` like so:
+
+```yaml
+enqueue:
+    default:
+        transport:            # Required
+        client:
+            traceable_producer:   true
+            prefix:               enqueue
+            separator:            .
+            app_name:             app
+            router_topic:         default
+            router_queue:         router # default is: default
+            router_processor:     null
+            redelivered_delay_time: 0
+            default_queue:        default
+```
+
+## Multiple topic queues
+
+Configuration for the different queues happens on the [processor](message_processor.md)/subscriber level - the queue is either defined in `config/services.yml` like so:
+
+```yaml
+# src/AppBundle/Resources/services.yml
+
+services:
+  AppBundle\Async\SayHelloProcessor:
+    tags:
+        # registers as topic processor
+        - { name: 'enqueue.processor', topic: 'hello', queue: 'say' }
+```
+
+or in the processor itself like so:
+
+```php
+<?php
+use Enqueue\Client\CommandSubscriberInterface;
+use Interop\Queue\Processor;
+
+class SayHelloProcessor implements Processor, TopicSubscriberInterface
+{
+    public static function getSubscribedTopics()
+    {
+        return [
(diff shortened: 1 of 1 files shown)


docs/bundle/multiple_queues.md
```diff
@@ -0,0 +1,70 @@
...
+{
+    public static function getSubscribedTopics()
+    {
+        return [
+           'topic' => 'hello',
+           'queue' => 'say',
+       ];
+    }
+}
+```
```
Steveb-p: I might be wrong here, but doesn't `TopicSubscriberInterface` work because it creates the above `tag` during autoconfiguration of Symfony `Container`? There was also a question regarding multiple `clients` (https://github.com/php-enqueue/enqueue-dev/issues/835) which might be mentioned here (specifically - disabling default tag and attaching to a specific client)

docs/bundle/multiple_queues.md
```diff
@@ -0,0 +1,70 @@
+---
+layout: default
+parent: "Symfony bundle"
+title: Functional testing
```
Steveb-p: You probably left the title here by accident.

docs/bundle/multiple_queues.md
```diff
@@ -0,0 +1,70 @@
...
+           'topic' => 'hello',
+           'queue' => 'say',
+       ];
+    }
+}
+```
+
+Both ways will enable you to run a queue called `say` exclusively with the [cli command](cli_commands.md)  `bin/console enqueue:consume say` or run all tasks except those in the `say` queue with `bin/console enqueue:consume --skip=say`
+
+Beware that messages for this topic will end up in the router queue first - so if you don't have a process running that consumes the router queue, your messages will never _really_ reach the `say` queue. In that case you'll still have to run `bin/console enqueue:consume default` - or, if you have changed the name of your router queue: `bin/console enqueue:consume router`
```
Steveb-p: We should probably provide a visual representation alongside this description how queues & topics interact. It might be outside the scope of this PR, but it might be worthwile in the long run?

docs/bundle/multiple_queues.md
lostfocus: Indeed. Will be fixed asap

docs/bundle/multiple_queues.md
lostfocus: I'd sure like to see that, too. :) My skills would be enough for a crude drawing in paint - I doubt that would be helpful for anyone.

docs/bundle/multiple_queues.md
lostfocus: You're not wrong - I just figured that the client use case already has an example in `message_processor.md`

docs/bundle/multiple_queues.md
```diff
@@ -0,0 +1,70 @@
...
```
makasim: The command [accepts  several queues](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue/Symfony/Client/ConsumeCommand.php#L84)

It would be good to suggest users use 

```
bin/console enqueue:consume say default
```

with clarifications on why default should be there 

Other than that it looks great! Thank you @lostfocus for your work.

docs/bundle/multiple_queues.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+or in the processor itself like so:
+
+```php
+<?php
+use Enqueue\Client\CommandSubscriberInterface;
+use Interop\Queue\Processor;
+
+class SayHelloProcessor implements Processor, TopicSubscriberInterface
```
makasim: fyi: you don't have to implement TopicSubscriberInterface interface if you use  `enqueue.processor` tag.

makasim: If you implement `TopicSubscriberInterface` then a tag is not needed as the service is autoconfigured with `enqueue.topic_subscriber` tag.

docs/bundle/multiple_queues.md
lostfocus: Life got into the way, sorry. :) I'll clear it up later.

lostfocus: Okay, I've re-read my paragraphs and I'm not all that sure why default should be there. I figured that with two queue runners - one only running the `say` queue and the other one running all and/or only the `default` default - everything should be fine?

balabis:
This is really missing from the documentation. Helped me a lot, thanks :)

Steveb-p:
Sorry @balabis, this PR kinda got lost :)

@lostfocus, would you mind revisiting this?

mleczakm:
This docs saved me a lot of time - it should be officially merged definitely!

PR #932, opened 2019-08-08
Add an extension to stop consumption on closed entity manager

Lctrs:


makasim:
for me, it would be better to stop consumption if an entity manager is closed.

Lctrs:
@makasim Failure seems unrelated.

Issue #894, opened 2019-06-08
Kafka: Offset commit (manual) failed


KonstantinCodes:
My consumer always fails to commit the offset:

```
%4|1560006005.731|COMMITFAIL|rdkafka#consumer-20| [thrd:main]: Offset commit (manual) failed for 1/1 partition(s): Broker: Unknown member: commands[0]@1(Broker: Unknown member)
```

Here is my Config:

```
        consumer:
            dsn: "..."
            global:
                ### Make sure this is unique for each application / consumer group and does not change
                ### Otherwise, Kafka won't be able to track your last offset and will always start according to
                ### `auto.offset.reset` setting.
                ### See Kafka documentation regarding `group.id` property if you want to know more
                group.id: 'backend'
                metadata.broker.list: '...'
                security.protocol: sasl_ssl
                ssl.ca.location: '%kernel.project_dir%/config/kafka/ca.pem'
                sasl.username: '...'
                sasl.password: '...'
                sasl.mechanisms: "SCRAM-SHA-256"
                max.poll.interval.ms: '25000'
                enable.auto.commit: 'false'
                enable.auto.offset.store: 'false'
            topic:
                auto.offset.reset: beginning
            ### Commit async is true by default since version 0.9.9.
            ### It is suggested to set it to true in earlier versions since otherwise consumers become extremely slow,
            ### waiting for offset to be stored on Kafka before continuing.
            commit_async: true
```

Any ideas, why this is happening?

Tested with both heroku Kafka & CloudKarafka.

Steveb-p:
What offset are you trying to commit?

I've looked up some issues in librdkafka and I've found:

> Are you trying to commit logical offsets (<0) ?
> Only absolute offsets (>= 0) will be committed.

Show sample code if possible. I've tried to see what the issue is, but it seems it's something between librdkafka & Kafka broker itself.

KonstantinCodes:
@Steveb-p Hi, progress has been slow, sorry for late answer.

I published it here: https://github.com/KonstantinCodes/messenger-kafka
I tried to make it as basic as possible and it works for me. In the bottom of the readme, I put the exact config which I'm using. And it runs. I updated the Ticket with the enqueue config, which produced the error to commit.

Hopefully it helps!

marcoreni:
[...]

Issue #893, opened 2019-06-08
Kafka: Application maximum poll interval exceeded


KonstantinCodes:
My consumer
```
        consumer:
            dsn: "..."
            global:
                ### Make sure this is unique for each application / consumer group and does not change
                ### Otherwise, Kafka won't be able to track your last offset and will always start according to
                ### `auto.offset.reset` setting.
                ### See Kafka documentation regarding `group.id` property if you want to know more
                group.id: 'backend'
                metadata.broker.list: '...'
                security.protocol: sasl_ssl
                ssl.ca.location: '%kernel.project_dir%/config/kafka/ca.pem'
                sasl.username: '...'
                sasl.password: '...'
                sasl.mechanisms: "SCRAM-SHA-256"
                max.poll.interval.ms: '25000'
                enable.auto.commit: 'false'
                enable.auto.offset.store: 'false'
            topic:
                auto.offset.reset: beginning
            ### Commit async is true by default since version 0.9.9.
            ### It is suggested to set it to true in earlier versions since otherwise consumers become extremely slow,
            ### waiting for offset to be stored on Kafka before continuing.
            commit_async: true
```

Is always producing these messages:
```
%4|1560006029.619|MAXPOLL|rdkafka#consumer-21| [thrd:main]: Application maximum poll interval (25000ms) exceeded by 333ms (adjust max.poll.interval.ms for long-running message processing): leaving group
%4|1560006054.153|MAXPOLL|rdkafka#consumer-23| [thrd:main]: Application maximum poll interval (25000ms) exceeded by 333ms (adjust max.poll.interval.ms for long-running message processing): leaving group
```

Do you have an idea, how to fix this?

Steveb-p:
@KonstantinCodes you're running a long running producer of some sort?

What phprdkafka / librdkafka combination you're running?

Overall, it is kind of related to #749 (not running `poll` specifically). They are warnings, so your application should work despite them (confirm?), but obviously we should do something about it.

KonstantinCodes:
@Steveb-p yeah it's kind of interesting.

I'm running a long running consumer, but the messages start popping up after 25 seconds.

So from what I read, you just have to run `KafkaConsumer::consume($timeout)`. Where the $timeout is smaller than your kafka config value `max.poll.interval.ms`.

[...]

PR #947, opened 2019-09-11
Kafka singleton consumer

dirk39:
Hi, introduce a `rdKafkaConsumers` collection to avoid the creation of multiple consumers for the same queue. This PR solve the issue #894 and #893  (I've tried the solution on this [project](https://github.com/marcoreni/enqueue-kafka-894-repro) and it works)

### DIFF
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -133,6 +133,7 @@ class RdKafkaContext implements Context
     {
         $kafkaConsumers = $this->kafkaConsumers;
         $this->kafkaConsumers = [];
+        $this->rdKafkaConsumers = [];
 
         foreach ($kafkaConsumers as $kafkaConsumer) {
             $kafkaConsumer->unsubscribe();
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -46,6 +46,11 @@ class RdKafkaContext implements Context
     private $kafkaConsumers;
 
     /**
+     * @var RdKafkaConsumer[]
+     */
+    private $rdKafkaConsumers;
+
+    /**
      * @param array $config
      */
     public function __construct(array $config)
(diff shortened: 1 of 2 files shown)


Steveb-p:
I was wondering if it's the responsibility of enqueue to keep consumers cached. Theoretically one should never want to create a second consumer to the same broker, but I could see a potential use-case when one creates two consumers with different `group.id`'s. You won't be able to do so if you're only checking for Kafka topic name.

I think this caching should be done in adapter instead. WDYT @dirk39 ?

dirk39:
@Steveb-p caching in adapter could be a good solution, but taking in examples the [sroze](https://github.com/sroze/messenger-enqueue-transport) implementation, he uses the same class (QueueInteropTransport) to execute the `ack` method and this is valid for any context. Kafka is sensible about consumers because he associate any consumer to a specific partition so i think is important to return same consumer if already instantiate.

makasim:
good to me. One note: clean cached consumers on context close.

pkg/rdkafka/RdKafkaContext.php
```diff
@@ -102,20 +107,26 @@ public function createConsumer(Destination $destination): Consumer
...
+                $this->getSerializer()
+            );
+
+            if (isset($this->config['commit_async'])) {
+                $consumer->setCommitAsync($this->config['commit_async']);
+            }
 
-        if (isset($this->config['commit_async'])) {
-            $consumer->setCommitAsync($this->config['commit_async']);
+            $this->rdKafkaConsumers[$queueName] = $consumer;
```
makasim: clean state when context is closed

dirk39:
Ok. @Steveb-p @makasim do you believe that could be useful to discriminate consumers by `group.id` and `queueName` or can we left just `queueName`?

Steveb-p:
> Ok. @Steveb-p @makasim do you believe that could be useful to discriminate consumers by `group.id` and `queueName` or can we left just `queueName`?

I believe using multiple consumers attached to the same topic with different `group.id`'s use-case is rather rare. I think we can leave it as is and come back to it if it's troublesome for anyone.

Theoretically in newer versions of librdkafka `group.id` is no longer necessary and can be used to force reading of a topic without storing offset, but that's another story.

pkg/rdkafka/RdKafkaContext.php
dirk39: done

pkg/rdkafka/RdKafkaContext.php
```diff
@@ -102,26 +107,33 @@ public function createConsumer(Destination $destination): Consumer
...
```
Steveb-p: A minor thing - `rdKafkaConsumers` property is not initialized in constructor.

dirk39:
Hi @Steveb-p , do you have an idea of when do you expect to merge the PR?

Steveb-p:
> Hi @Steveb-p , do you have an idea of when do you expect to merge the PR?

I'm only waiting for confirmation from @makasim.

Steveb-p:
@makasim ?

makasim:
There are two properties `kafkaConsumers` and `rdkafkaConsumers`. I could not spot a difference.

Steveb-p:
> There are two properties `kafkaConsumers` and `rdkafkaConsumers`. I could not spot a difference.

@makasim `kafkaConsumers` keeps references to enqueue-based classes wrapping rdkafka consumers, while `rdkafkaConsumers` are rdkafka consumers without those wrappers. Wrappers do not allow direct access to internal consumers, so that's why they are separate.

Otherwise we should introduce a getter method to wrappers (which is fine and probably should be prefered?)

makasim:
> Otherwise we should introduce a getter method to wrappers (which is fine and probably should be prefered?)

It could be done as a separate PR

PR #955, opened 2019-09-25
Add header support for kafka

TiMESPLiNTER:
This bumps the dependency on rdkafka extension from version `3.0.3` to `3.1.0`.

### DIFF
--- a/pkg/rdkafka/composer.json
+++ b/pkg/rdkafka/composer.json
@@ -7,7 +7,7 @@
     "license": "MIT",
     "require": {
         "php": "^7.1.3",
-        "ext-rdkafka": "^3.1.0",
+        "ext-rdkafka": "^3.0.3",
         "queue-interop/queue-interop": "^0.7|^0.8"
     },
     "require-dev": {
--- a/pkg/rdkafka/composer.json
+++ b/pkg/rdkafka/composer.json
@@ -7,7 +7,7 @@
     "license": "MIT",
     "require": {
         "php": "^7.1.3",
-        "ext-rdkafka": "^3.0.3",
+        "ext-rdkafka": "^3.1.0",
         "queue-interop/queue-interop": "^0.7|^0.8"
     },
     "require-dev": {
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -138,6 +138,17 @@ class RdKafkaContext implements Context
         throw PurgeQueueNotSupportedException::providerDoestNotSupportIt();
     }
 
+    public static function getLibrdKafkaVersion(): string
+    {
+        if (!defined('RD_KAFKA_VERSION')) {
+            throw new \RuntimeException('RD_KAFKA_VERSION constant is not defined. Phprdkafka is probably not installed');
+        }
+        $major = (RD_KAFKA_VERSION & 0xFF000000) >> 24;
+        $minor = (RD_KAFKA_VERSION & 0x00FF0000) >> 16;
+        $patch = (RD_KAFKA_VERSION & 0x0000FF00) >> 8;
+        return "$major.$minor.$patch";
+    }
+
     private function getProducer(): VendorProducer
     {
         if (null === $this->producer) {
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -146,6 +146,7 @@ class RdKafkaContext implements Context
         $major = (RD_KAFKA_VERSION & 0xFF000000) >> 24;
         $minor = (RD_KAFKA_VERSION & 0x00FF0000) >> 16;
         $patch = (RD_KAFKA_VERSION & 0x0000FF00) >> 8;
+
         return "$major.$minor.$patch";
     }
 
--- a/pkg/rdkafka/RdKafkaProducer.php
+++ b/pkg/rdkafka/RdKafkaProducer.php
@@ -50,12 +50,13 @@ class RdKafkaProducer implements Producer
             if (version_compare(RdKafkaContext::getLibrdKafkaVersion(), '1.0.0', '>=')
                 && version_compare(phpversion('rdkafka'), '3.1.0', '<=')) {
                 trigger_error(
-                    'Phprdkafka <= 3.1.0 is incompatible with librdkafka 1.0.0 when calling `producev`. ' .
+                    'Phprdkafka <= 3.1.0 is incompatible with librdkafka 1.0.0 when calling `producev`. '.
                     'Falling back to `produce` (without message headers) instead.',
                     E_USER_WARNING
                 );
             } else {
                 $topic->producev($partition, 0 /* must be 0 */, $payload, $key, $message->getHeaders());
+
                 return;
             }
         }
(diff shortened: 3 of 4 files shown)


makasim:
looks good to me. @Steveb-p wdyt?

TiMESPLiNTER:
I get the following (I guess) non-related error running the functional tests:

``` 
1) Enqueue\Dbal\Tests\Spec\Mysql\DbalSubscriptionConsumerConsumeFromAllSubscribedQueuesTest::test

Failed asserting that two arrays are equal.

--- Expected

+++ Actual

@@ @@

 Array (

     0 => 'fooBody'

-    1 => 'barBody'

 )

/mqdev/vendor/queue-interop/queue-spec/src/SubscriptionConsumerConsumeFromAllSubscribedQueuesSpec.php:63
```

Steveb-p:
I was sure I did open a PR earlier with it: #880. Closed it since a lot has changed.

However, I'd suggest getting the part with backward compatibility ported here?

```php
// Merge headers passed from Kafka with possible earlier serialized payload headers. Prefer Kafka's.
// Note: Requires phprdkafka >= 3.1.0
if (isset($kafkaMessage->headers)) {
    $message->setHeaders(array_merge($message->getHeaders(), $kafkaMessage->headers));
}
```
This way messages produced by earlier versions of `php-enqueue` would still be handled.

I'll give it a look after work, but it should be fine :)

TiMESPLiNTER:
So should I just copy over the code from your closed PR into this PR?

```php
        // Note: Topic::producev method exists in phprdkafka > 3.1.0
        // Headers in payload are maintained for backwards compatibility with apps that might run on lower phprdkafka version
        if (method_exists($topic, 'producev')) {
            // Phprdkafka <= 3.1.0 will fail calling `producev` on librdkafka 1.0.0 causing segfault
            if (version_compare(RdKafkaContext::getLibrdKafkaVersion(), '1.0.0', '>=')
                && version_compare(phpversion('rdkafka'), '3.1.0', '<=')) {
                trigger_error('Phprdkafka < 3.1.0 is incompatible with librdkafka 1.0.0 when calling `producev`', E_USER_WARNING);
            }
            $topic->producev($partition, 0 /* must be 0 */, $payload, $key, $message->getHeaders());
        } else {
            $topic->produce($partition, 0 /* must be 0 */, $payload, $key);
        }
```

Steveb-p:
> So should I just copy over the code from your closed PR into this PR?
> 
> ```
>         // Note: Topic::producev method exists in phprdkafka > 3.1.0
>         // Headers in payload are maintained for backwards compatibility with apps that might run on lower phprdkafka version
>         if (method_exists($topic, 'producev')) {
>             // Phprdkafka <= 3.1.0 will fail calling `producev` on librdkafka 1.0.0 causing segfault
>             if (version_compare(RdKafkaContext::getLibrdKafkaVersion(), '1.0.0', '>=')
>                 && version_compare(phpversion('rdkafka'), '3.1.0', '<=')) {
>                 trigger_error('Phprdkafka < 3.1.0 is incompatible with librdkafka 1.0.0 when calling `producev`', E_USER_WARNING);
>             }
>             $topic->producev($partition, 0 /* must be 0 */, $payload, $key, $message->getHeaders());
>         } else {
>             $topic->produce($partition, 0 /* must be 0 */, $payload, $key);
>         }
> ```

~~I'll be ok with enforcing higher level of phprdkafka/librdkafka instead and keeping your code that changes this to call `producev` by default.~~

Yes actually. That would result in still allowing to work with lower phprdkafka versions, while using the new `producev` method when available. There is a lot of recent contributions to phprdkafka (Thanks to Nick), but there are also reports about some segfaults happening sometimes, so I'm not keen to forcing usage of those versions yet (but that's only my opinion).

My original code should probably be changed to throw an `Exception` or trigger a full error instead, because phprdkafka 3.1.0 will simply not `producev` message properly with librdkafka `1.0.0` (segfault was guaranteed to happen). Or it should drop to `produce` call instead. WDYT @TiMESPLiNTER ?

TiMESPLiNTER:
@Steveb-p how about dropping to produce and trigger a PHP warning in the 3.1.0/1.0.0 case?

Steveb-p:
@TiMESPLiNTER :+1:

TiMESPLiNTER:
@Steveb-p @makasim I added the functionality as discussed. Please have a look over the code again if you find some time. :smiley:

Steveb-p:
> I think we could include this in the next bugfix release for `0.9.x`? What would I need to do to include it there? Also make a PR against the `0.9` branch?

`0.9` branch afair is actually a "tag" for initial `0.9` release ;)

`master` branch is the base for new releases, so it's fine as it is.

Steveb-p:
@TiMESPLiNTER would you mind adding 

https://github.com/php-enqueue/enqueue-dev/pull/880/files#diff-30b559bffa4a02ebf46fbef2767f7a01R175-R180

as well?

It would provide cross-compatibility with earlier versions (which might actually be important since Kafka can store all historical messages, right? :smile: )

And the inclusion of that code for `produce` fallback means that version from `^3.0.3` to `^3.1.0` is no longer required.

pkg/rdkafka/composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "MIT",
     "require": {
         "php": "^7.1.3",
-        "ext-rdkafka": "^3.0.3",
+        "ext-rdkafka": "^3.1.0",
```
Steveb-p: > And the inclusion of that code for `produce` fallback means that version from `^3.0.3` to `^3.1.0` is no longer required.

TiMESPLiNTER: Oops, forgot about it. Of course you're right. Reverted it.

TiMESPLiNTER:
> @TiMESPLiNTER would you mind adding
> 
> https://github.com/php-enqueue/enqueue-dev/pull/880/files#diff-30b559bffa4a02ebf46fbef2767f7a01R175-R180
> 
> as well?

This code is currently already in master from what I see. Just the producing functionality is currently missing header support.

Steveb-p:
@makasim IMO it's ready. As usual I'm not merging it since you're the owner :smiling_imp:

TiMESPLiNTER:
When is the next released planned (including this "bug fix")?

Steveb-p:
> When is the next released planned (including this "bug fix")?

Until @makasim creates a new version (which should be soon™) you can use this functionality of composer to force it to use your fork:

> There are a few use cases for this. The most common one is maintaining your own fork of a third party library. If you are using a certain library for your project and you decide to change something in the library, you will want your project to use the patched version. If the library is on GitHub (this is the case most of the time), you can simply fork it there and push your changes to your fork. After that you update the project's composer.json. All you have to do is add your fork as a repository and update the version constraint to point to your custom branch. In composer.json, you should prefix your custom branch name with "dev-". For version constraint naming conventions see Libraries for more information.

So, in your case it might be:

```json
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/TiMESPLIiNTER/enqueue/rdkafka"
        }
    ],
    "require": {
        "enqueue/rdkafka": "^0.9.13 | dev->>BRANCH-NAME<<"
    }
}
```

Adding `^0.9.13` will allow you to update to a newer version once it's available (by using `composer update enqueue/rdkafka`), if you have `prefer-stable:true` in `composer.json` config (https://getcomposer.org/doc/04-schema.md#prefer-stable)

makasim:
Ths fix already in enqueue/rdkafka, you require dev version for now. 

I'll do release when I have time

TiMESPLiNTER:
Okay, just was wondering if something is planned. All good I'll use `dev-master` in the meantime.

PR #956, opened 2019-09-26
New message transformation concept

TiMESPLiNTER:
This PR aims to deprecate the current  `Serializer` approach and replaces it with a "message transformation concept" which allows to modify the message before after consumption before sending. (For example to de/encode the payload if it's in Avro format.)

This PR should not introduce any BC.

### DIFF
--- a/pkg/rdkafka/ConsumeMessageTransformer.php
+++ b/pkg/rdkafka/ConsumeMessageTransformer.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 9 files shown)


Steveb-p:
@TiMESPLiNTER Actually you might want to look into #846 and PRs/issues linked in it's description.

Long story short there was a discussion about something similar to what you're proposing, but with the intention of allowing passing arbitrary types as messages to allow custom made serialization on any desired level. 

I'll look into your PR further when I've got spare time if you dont mind :smile:

PR #958, opened 2019-10-04
Fix empty class for autowired services (Fix #957)

### LINKED ISSUE #957: Empty class from the definition of the service
Hello there !

I noticed a small bug about service definition in your bundle. If I declare an autowired service from Symfony 3.4, I **must** set the `class` parameter because of the `BuildTopicSubscriberRoutesPass` that needs it.

In fact, when autowiring/autoconfiguring services actually works, but the `class` value of the definition is set to `null` if not explicitly declared.

Here is a small example:

```yaml
App\Bundle\RestBundle\Processor\GalaxyRecordProcessor:
    # class: App\Bundle\RestBundle\Processor\GalaxyRecordProcessor
    tags: { name: enqueue.topic_subscriber }
```

that produces the following error:

```
In BuildTopicSubscriberRoutesPass.php line 42:
                                              
  The processor class "" could not be found.
```

So the `CompilerPass` tries to get an empty class because it does not understand that the service name is actually the class. A small fix would be to replace this line:

```php
$processorClass = $processorDefinition->getClass();
```

by this:

```php
$processorClass = $processorDefinition->getClass() ?? $serviceId;
```

I'll make a PR as soon as possible (Oh my god ! That `docker-compose up -d` is killing me ! :p)

PS: Same thing for `BuildCommandSubscriberRoutesPass` by the way ;)

Cheers guys !

NicolasGuilloux:
Fix empty class for autowired services at least in Symfony 3.4 (Not tested for other versions).

More information in the related issue #957.

### DIFF
--- a/pkg/enqueue/Symfony/Client/DependencyInjection/BuildCommandSubscriberRoutesPass.php
+++ b/pkg/enqueue/Symfony/Client/DependencyInjection/BuildCommandSubscriberRoutesPass.php
@@ -35,7 +35,7 @@ final class BuildCommandSubscriberRoutesPass implements CompilerPassInterface
                     throw new \LogicException('The command subscriber tag could not be applied to a service created by factory.');
                 }
 
-                $processorClass = $processorDefinition->getClass();
+                $processorClass = $processorDefinition->getClass() ?? $serviceId;
                 if (false == class_exists($processorClass)) {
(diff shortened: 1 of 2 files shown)


NicolasGuilloux:
I did not edited the tests now. If somebody has advices, let me know :)

makasim:
Thanks

PR #959, opened 2019-10-04
Compatibility with Phprdkafka 4.0

Steveb-p:
@see https://github.com/arnaud-lb/php-rdkafka/releases/tag/4.0.0

Since `poll` no longer gets called on destruct/shutdown, we should provide error handling (#749).

I'm marking this as draft PR since otherwise we might have people losing messages when PHP process shuts down before message is acknowledged by Kafka broker.

### DIFF
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -36,7 +36,7 @@ class RdKafkaContext implements Context
     private $conf;
 
     /**
-     * @var Producer
+     * @var RdKafkaProducer
      */
     private $producer;
 
@@ -96,7 +96,23 @@ class RdKafkaContext implements Context
      */
     public function createProducer(): Producer
     {
-        return new RdKafkaProducer($this->getProducer(), $this->getSerializer());
+        if (!isset($this->producer)) {
+            $producer = new VendorProducer($this->getConf());
+
+            if (isset($this->config['log_level'])) {
+                $producer->setLogLevel($this->config['log_level']);
+            }
+
+            $this->producer = new RdKafkaProducer($producer, $this->getSerializer());
+
+            // Once created RdKafkaProducer can store messages internally that need to be delivered before PHP shuts
+            // down. Otherwise, we are bound to lose messages in transit.
+            // Note that it is generally preferable to call "close" method explicitly before shutdown starts, since
+            // otherwise we might not have access to some objects, like database connections.
+            register_shutdown_function([$this->producer, 'flush'], $this->config['shutdown_timeout'] ?? -1);
+        }
+
+        return $this->producer;
     }
 
     /**
@@ -139,6 +155,11 @@ class RdKafkaContext implements Context
         foreach ($kafkaConsumers as $kafkaConsumer) {
             $kafkaConsumer->unsubscribe();
         }
+
+        // Compatibility with phprdkafka 4.0.
+        if (isset($this->producer)) {
+            $this->producer->flush($this->config['shutdown_timeout'] ?? -1);
+        }
     }
 
     public function createSubscriptionConsumer(): SubscriptionConsumer
@@ -161,19 +182,6 @@ class RdKafkaContext implements Context
         $patch = (RD_KAFKA_VERSION & 0x0000FF00) >> 8;
 
         return "$major.$minor.$patch";
-    }
-
-    private function getProducer(): VendorProducer
-    {
-        if (null === $this->producer) {
-            $this->producer = new VendorProducer($this->getConf());
-
-            if (isset($this->config['log_level'])) {
-                $this->producer->setLogLevel($this->config['log_level']);
-            }
-        }
-
-        return $this->producer;
     }
 
     private function getConf(): Conf
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -50,9 +50,6 @@ class RdKafkaContext implements Context
      */
     private $rdKafkaConsumers;
 
-    /**
-     * @param array $config
-     */
     public function __construct(array $config)
     {
         $this->config = $config;
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ script:
     - if [ "$PHPSTAN" = true ] && [ ! -z "${PKG_PHP_CHANGED_FILES}" ]; then docker run --workdir="/mqdev" -v "`pwd`:/mqdev" --rm enqueue/dev:latest php -d memory_limit=1024M bin/phpstan analyse -l 1 -c phpstan.neon -- ${PKG_PHP_CHANGED_FILES[@]} ; fi
     - if [ "$UNIT_TESTS" = true ]; then bin/phpunit --exclude-group=functional; fi
     - if [ "$FUNCTIONAL_TESTS" = true ]; then bin/test.sh --exclude-group=rdkafka; fi
-    - if [ "RDKAFKA_TESTS" = true ]; then bin/test.sh --group=rdkafka; fi
+    - if [ "$RDKAFKA_TESTS" = true ]; then bin/test.sh --group=rdkafka; fi
 
 notifications:
   webhooks:
--- a/composer.json
+++ b/composer.json
@@ -59,7 +59,7 @@
     "doctrine/doctrine-bundle": "~1.2|^2",
     "doctrine/mongodb-odm-bundle": "^3.5|^4",
     "alcaeus/mongo-php-adapter": "^1.0",
-    "kwn/php-rdkafka-stubs": "^1.0.2",
+    "kwn/php-rdkafka-stubs": "^1.0.2 | ^2.0",
     "friendsofphp/php-cs-fixer": "^2"
   },
   "autoload": {
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -99,13 +99,13 @@ services:
       - '2181:2181'
 
   kafka:
-    image: 'wurstmeister/kafka:0.10.2.1'
+    image: 'wurstmeister/kafka'
     ports:
       - '9092:9092'
     environment:
       KAFKA_ZOOKEEPER_CONNECT: 'zookeeper:2181'
-    volumes:
-      - '/var/run/docker.sock:/var/run/docker.sock'
+      KAFKA_ADVERTISED_HOST_NAME: kafka
+      KAFKA_ADVERTISED_PORT: 9092
 
   google-pubsub:
       image: 'google/cloud-sdk:latest'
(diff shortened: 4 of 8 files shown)


anam-hossain:
@Steveb-p What are the code changes required for handling this issue? Do you have any example? We are facing the same issue. Thanks in advanced.

Steveb-p:
@anam-hossain #749 needs to be resolved.

I've seen your response about `flush` in phprdkafka repository, I'll try to work on it soon™.

Steveb-p:
@TiMESPLiNTER @makasim @nick-zh if you could take a look at this PR and see if I'm doing it alright. This should roughly replicate the current behavior for phprdkafka 3.x and wait for message delivery if there is no special configuration present (`shutdown_timeout`).

There is still the "issue" of `setDefaultTopicConf` being deprecated (btw @nick-zh, it might be worth adding it to documentation of phprdkafka?). What is the expected way of handling this? How `TopicConf` should be delivered to an instance?

pkg/rdkafka/RdKafkaContext.php
```diff
@@ -139,6 +139,12 @@ public function close(): void
         foreach ($kafkaConsumers as $kafkaConsumer) {
             $kafkaConsumer->unsubscribe();
         }
+
+        // Compatibility with phprdkafka 4.0.
+        if (isset($this->producer) && method_exists($this->producer, 'flush')) {
+            $this->producer->flush($this->config['shutdown_timeout'] ?? -1);
```
nick-zh: maybe throw an error or log something if the return value is not `RD_KAFKA_RESP_ERR_NO_ERROR`, sry not sure what the enqueue convention is for something like that

Steveb-p: As what we discussed: throwing an exception here would change the behavior and make it necessary for wrapping libraries (like enqueue-bundle) to catch it, so I've left it out.

However, I do agree that it is something that this package eventually should do, since otherwise potential delivery errors will be hidden from end user.

Steveb-p:
Issues we're steming from changes in image Dockerfile is based on (PHP version changed to 7.3) which caused phprdkafka extension to not be loaded and stubs were being used instead confused. I'd say we drop stubs loading in bootstrap.php in this situation? It's only really relevant when doing unit tests and it shouldn't really be that way - it makes those classes behave like phpunit's mocks :/

I've streamlined Dockerfile, spliting especially loading external OS dependencies for build and sorting them alphabetically.

nick-zh:
Except for the error handling, i think the changes do look ok.
Regarding the default topic conf @Steveb-p , all the configs that were set there, can be set in the normal config object.

pkg/rdkafka/RdKafkaContext.php
```diff
@@ -171,6 +177,12 @@ private function getProducer(): VendorProducer
             if (isset($this->config['log_level'])) {
                 $this->producer->setLogLevel($this->config['log_level']);
             }
+
+            // Once created RdKafkaProducer can store messages internally that need to be delivered before PHP shuts
+            // down. Otherwise, we are bound to lose messages in transit.
+            // Note that it is generally preferable to call "close" method explicitly before shutdown starts, since
+            // otherwise we might not have access to some objects, like database connections.
+            register_shutdown_function([$this, 'close']);
```
maks-rafalko: looks like this will be called as many times as `getProduce()` is executed. 

I would suggest to add an early return or something here to call shutdown function only once

docs/contribution.md
```diff
@@ -13,9 +13,25 @@ The pull requests to read only subtree split [repositories](https://github.com/p
...
+PR's go through automatic testing and code style checks. For your convenience 3 commands below will set up your local
+environment. You will need:
+1. A globally accessible `composer` script for package dependencies
+2. Docker & `docker-compose` for running tests against services
+
+Note it is assumed your `docker-compose` is available without sudo.
+
+First install dependencies:
+```
+composer install --ignore-platform-reqs
```
maks-rafalko: Could you please explain why would we need `--ignore-platform-reqs`? Looks dangerous for me

Steveb-p: It's only a suggestion for local development in cases where not every extension is installed (for example, you might be missing `ext-mongo` or `ext-rdkafka`). Otherwise you're forced to install theoretically unrelated extensions, even if you only intend to work on a specific subset of enqueue-dev functionality.

maks-rafalko: this is what happens when the project is not split by different packages with carefully thought dependencies :(

pkg/rdkafka/RdKafkaContext.php
Steveb-p: > looks like this will be called as many times as `getProduce()` is executed.

This is kept inside an `if` statement that only initializes if `$this->producer` has not yet been set, and it's the only place where this property is set. So it will called at most once for each context.

maks-rafalko: I see, didn't notice parent `if`. Thank you

docs/contribution.md
Steveb-p: It's a concept of monorepo that is not really uncommon. We can discuss it's merits and demerits, but you can encounter it in a couple of different open source projects, like [Symfony](https://github.com/symfony/symfony/blob/master/composer.json).

In this case, it allows all packages to inherit the main structure and be tested inside one, main project. This makes cross-transport changes **a lot** easier.

Eventually, subpackages are split to subrepositories for consumption in packagist.

makasim: In fact, you don't need `--ignore-platform-reqs`. enqueue-dev use compose platforms feature to install deps and runs tests in prepared containers where all extensions and services are in place. 

Split repos have correct deps too

Steveb-p: > In fact, you don't need `--ignore-platform-reqs`. enqueue-dev use compose platforms feature to install deps and runs tests in prepared containers where all extensions and services are in place.

@makasim true, but only to an extent - contribution docs specifically say what has to be done to get enqueue-dev to install dependencies for local, non-containerized environment. That's why I suggested to add `--ignore-platform-reqs` for it, in cases where someone wants to work on a specific transport and doesn't really want to install other extensions that he/she doesn't care about.

I could convert it into a note instead or drop it entirely, working on assumption that developer will have enough knowledge to handle it himself.

makasim: This https://github.com/php-enqueue/enqueue-dev/blob/master/composer.json#L110-L120 should do for local setup.

Steveb-p: Ouch. For some reason my local setup had a lock file that seemed to not include those platform options :disappointed: I'll remove it.

Steveb-p:
I've revisited the branch and removed all changes unrelated to actual compatibility with phprdkafka 4.0.

Now all this does is introduce `shutdown_timeout` configuration property that takes effect when working with phprdkafka 4.0.

Other changes will be opened as separate PRs.

nobYsDarling:
@makasim are there any contributions needed to merge this?

Egari:
I would also very much like to use enqueue/rdkafka on PHP 7.4 with phprdkafka 4.

@Steveb-p @makasim any prospect on seeing this PR reviewed and/or merged?

makasim:
> @makasim are there any contributions needed to merge this?

I've no idea. 

@Steveb-p is it good to go?

Steveb-p:
@makasim It should, I think it was working last time I've worked with it. But a lot of time has passed and I'm not sure - at the very least I'll need to take a look at it again.

makasim (approved):

makasim:
Looks good to me.

pkg/rdkafka/RdKafkaContext.php
```diff
@@ -96,7 +93,23 @@ public function createTemporaryQueue(): Queue
      */
     public function createProducer(): Producer
     {
-        return new RdKafkaProducer($this->getProducer(), $this->getSerializer());
+        if (!isset($this->producer)) {
+            $producer = new VendorProducer($this->getConf());
+
+            if (isset($this->config['log_level'])) {
+                $producer->setLogLevel($this->config['log_level']);
```
nick-zh, line 100: if this `setLogLevel` of `RdKafka\Producer` be aware that it has been deprecated

Steveb-p: @nick-zh you're correct, this is directly calling `setLogLevel` on Kafka Producer instance.

At this point I'm not really going to remove this deprecation, since in general PHP frameworks and applications should only log the deprecation (similarly to the default topic case). I'll address this in a separate PR.

Steveb-p:
@makasim I'd say it's good to go.

There is a side effect of shutdown function being registered that mimics pre-4.0 kafka extension behavior, but I think it's... acceptable? For 3.0 versions it will be simply a no-op since `flush` does not exist.

nikolaposa:
Hi @Steveb-p, 👋 sorry for awaking this PR. Do you know what's the preferred value for `shutdown_timeout` configuration for production environment? I see that the default value is `-1`, which I guess means "no timeout"?

Also, in general, do you have recommendations for low-latency Producer configuration? In the php-rdkafka docs I've found this: https://github.com/arnaud-lb/php-rdkafka#performance--low-latency-settings, but I'm wondering if there's more? What I had on mind was something like this:

```
socket.timeout.ms: 50
queue.buffering.max.ms: 1
connection.max.idle.ms: 1000
topic.metadata.refresh.interval.ms: 500
```

I'm also wondering whether `queue.buffering.max.ms: 1` means that there's no need for explicit `flush()` since there will be no batching of messages.

Thanks for all the great work you did here! 👍

nick-zh:
Hey @nikolaposa 

These two links should help [latency 1](https://github.com/edenhill/librdkafka/wiki/How-to-decrease-message-latency) and [latency 2](https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#low-latency) from the c library.

nikolaposa:
Thank you @nick-zh. 👍 Is `flush()` automatically called with default setup of this library? I see this `register_shutdown_function` call https://github.com/php-enqueue/enqueue-dev/pull/959/files#diff-3e3dc85ef34c5e358f190d3af6de75fc1f81596044f068d0cddc6d916d029c32R109. Are there any side effects if it is called more than once in a single request? For example, manually and then in register_shutdown_function?

nick-zh:
No worries, yeah sry i forgot about this, seems it will flush when you call `close()`
Regarding side effects, yes and no, if you are only producing one message no, the second flush will be instant.
If you produce more than one message, it will increase delay if you call it after each message.
Many settings / calls depend on your use case and what you try to achieve (how many messages, accepted delay time, etc.)
If you are just sending one message (e.g. triggered by a rest request), then not much optimization besides maybe shutdown is needed. It is a whole other story if it is a long running process with many messages, then the settings that i linked will help.
I hope that helps :smile:

nikolaposa:
Thank you once again Nick, that's very useful to know.

felippeduarte:
Hey. Looking at the rdkafka docs (https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka.flush.html)  flush returns an `integer` instead of `void` in Producer class. Can we change that? Will make it easier for some processes here that required the return code.

PR #965, opened 2019-10-10
Add header support for Symfony's produce command

### LINKED ISSUE #963: Add options to send headers with enqueue:produce command
There is currently no possibility to send headers with the  `enqueue:produce` command. It would be nice to add an option to specify headers:

```
$ bin/console enqueue:produce '{"foo": "message content"}' --topic=... --header='Content-Type: application/json' --header='Action: create' 
```

TiMESPLiNTER:
Fixes #963

### DIFF
--- a/pkg/enqueue/Symfony/Client/ProduceCommand.php
+++ b/pkg/enqueue/Symfony/Client/ProduceCommand.php
@@ -2,6 +2,7 @@
 
 namespace Enqueue\Symfony\Client;
 
+use Enqueue\Client\Message;
 use Enqueue\Client\ProducerInterface;
 use Psr\Container\ContainerInterface;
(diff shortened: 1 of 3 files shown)


Steveb-p:
@TiMESPLiNTER You can check local test run script instructions here https://github.com/php-enqueue/enqueue-dev/blob/master/docs/contribution.md (kinda obscure, that's why I'm linking it in case you've missed it)

Steveb-p:
Looks good to me. I've double checked that `SimpleClient` does indeed automatically convert string messages into `Message` instance.

Steveb-p:
@makasim Looks simple enough to me. Could I merge PR's like this in the future in case you're occupied or you'd like to keep control over it?

makasim:
@Steveb-p sure, I don't mind

PR #972, opened 2019-10-21
[amqp-lib] fast unsubscribe, do not use internal properties

ramunasd:
* do not wait for unsubscribe response
* remove usage of internal AMQPChannel property `callbacks`, they are handled internally
* style fixes


ramunasd:
careless bot..

Issue #981, opened 2019-11-20
[kafka] Useless while?


TiMESPLiNTER:
I have a question regarding these lines in the kafka consumer:

https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/RdKafkaConsumer.php#L103-L111

```php
        if ($timeout > 0) {
            $message = $this->doReceive($timeout);
        } else {
            while (true) {
                if ($message = $this->doReceive(500)) {
                    break;
                }
            }
        }
``` 

just out of curiosity: why isn't this written like this?

```php
        if ($timeout > 0) {
            $timeout = 500;
        }

        $message = $this->doReceive($timeout);
```

Steveb-p:
It's done to allow indefinite wait for message (when `$timeout` is 0). AFAIR phprdkafka always requires a timeout for consumption call.

It could be replaced with early return though, and the same applies to `doReceive` method.

Steveb-p:
@TiMESPLiNTER this is even more relevant if you consider possible callbacks set for - for example - rebalancing operations on the underlying consumer: https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka-kafkaconsumer.consume.html.

Note that there are two distinct operations - one with `timeout` explicitly set, and second when timeout was either set to 0 or not passed at all.

TiMESPLiNTER:
> It's done to allow indefinite wait for message (when `$timeout` is 0). AFAIR phprdkafka always requires a timeout for consumption call.

That's true. That's why I suggest to set it to `500` if you pass `0`. It removes the busy wait which imo is never a good option. For example if you use it with symfony messenger und you consume multiple (kafka) transports with the same messenger:consume command it blocks forever this very transport and never consumes from others. I also cannot imagine any use case where you need a forced busy wait. If you just consume from one topic and `0`  timeout defaults to `500` its anyway the same as the while solution currently in place. Or am I wrong?

Steveb-p:
> If you just consume from one topic and `0` timeout defaults to `500` its anyway the same as the while solution currently in place.

It doesn't do that.

If `$timeout` argument is set to 0, then `receive` method is turned into a blocking method - that's the while part. It will constantly try to get a message until it does.  
Counterpart to that is passing a `$timeout` value higher than 0. In this case it will pass **exactly** this value to `KafkaConsumer`.

[...]

PR #982, opened 2019-11-20
Prefer early returns in consumer code

Steveb-p:
Related to #981


Steveb-p:
@TiMESPLiNTER could you review this if it's to your liking? :smile:

PR #985, opened 2019-11-22
#984 Symfony 4.4 compatibility

ctrl-f5:
Fix fox issue #984 and Symfony 4.4 compatibility

see https://github.com/symfony/symfony/issues/32394

PR #988, opened 2019-11-23
Fix Incompatibility for doctrine

Baachi:
Hey guys,

i adjusted the code from #985 and added a test for this class.

#SymfonyHackday

### DIFF
--- a/pkg/enqueue/Doctrine/DoctrineConnectionFactoryFactory.php
+++ b/pkg/enqueue/Doctrine/DoctrineConnectionFactoryFactory.php
@@ -2,7 +2,7 @@
 
 namespace Enqueue\Doctrine;
 
-use Doctrine\Bundle\DoctrineBundle\Registry;
+use Doctrine\Common\Persistence\ManagerRegistry;
 use Enqueue\ConnectionFactoryFactoryInterface;
 use Enqueue\Dbal\ManagerRegistryConnectionFactory;
 use Enqueue\Dsn\Dsn;
@@ -11,7 +11,7 @@ use Interop\Queue\ConnectionFactory;
 class DoctrineConnectionFactoryFactory implements ConnectionFactoryFactoryInterface
 {
     /**
-     * @var Registry
+     * @var ManagerRegistry
      */
     private $doctrine;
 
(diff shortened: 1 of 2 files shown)


rogamoore:
👍 Anything preventing from merging and releasing this fix? :)

Steveb-p:
@rogamoore I'm usually not the one merging, since I've opted to take care Kafka part of Enqueue, but it seems @makasim is recently pretty busy.

I'll let it "bake" for some time, in case someone has any additional comments (like I've made the assumption that previous PR doing this is OK while it was pointing to wrong class, so I don't want to hastily merge).

At the same time, you can use Composer's functionality to overwrite packages to temporarily provide your own fork instead of a package: https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository

> Loading a package from a VCS repository
There are a few use cases for this. The most common one is maintaining your own fork of a third party library. If you are using a certain library for your project and you decide to change something in the library, you will want your project to use the patched version. If the library is on GitHub (this is the case most of the time), you can simply fork it there and push your changes to your fork. After that you update the project's composer.json. All you have to do is add your fork as a repository and update the version constraint to point to your custom branch. In composer.json, you should prefix your custom branch name with "dev-".

makasim:
It looks good. One thing though: Doctrine version has to be updated in composer.json. 

@Steveb-p I don't use the library nor work with PHP these days. Feel free to take over it (if you have that desire of course).

Steveb-p:
@makasim I can take it over if you're not working with it anymore. As we discussed in another issue (the one where you were looking for maintainers) I'll keep directly maintaining only the part with Kafka, since that's the package I'm familiar with, but will take a little more active approach with PR's from other packages.

I'll check/review this PR in more detail tonight after work @rogamoore

makasim:
Thank you @Steveb-p

rogamoore:
Regarding the Doctrine version I think this fix is still compatible with Doctrine Bundle <2.0 because it's just an alias that was removed, so using the `Doctrine\Common\Persistence\ManagerRegistry` directly should have been possible already before.

alcaeus:
> Regarding the Doctrine version I think this fix is still compatible with Doctrine Bundle <2.0 because it's just an alias that was removed, so using the `Doctrine\Common\Persistence\ManagerRegistry` directly should have been possible already before.

That is correct - this uses the base interface for all manager registries.

Issue #837, opened 2019-05-04
SQS Message ID


tigerstrikemedia:
The SQS Message handler does not set the AWS `messageID` to the message ID can we add

```
$message->setMessageId( $sqsMessage['MessageId'] );
```

to the `SqsConsumer` `convertMessage()` method ?

Thanks
Ben

makasim:
LGTM

makasim:
Feel free to send us a PR that address the issue

PR #992, opened 2019-11-29
[SQS] add messageId to the sqsMessage

BenoitLeveque:
Hi,

I saw that we are not allowing user to access the messageId of a sqs message.

following the issue #837 i have added the messageId in the convertMessage method.

### DIFF
--- a/pkg/sqs/SqsConsumer.php
+++ b/pkg/sqs/SqsConsumer.php
@@ -185,6 +185,10 @@ class SqsConsumer implements Consumer
     {
         $message = $this->context->createMessage();
 
+        if (isset($sqsMessage['MessageId'])) {
+            $message->setMessageId($sqsMessage['MessageId']);
+        }
+
         $message->setBody($sqsMessage['Body']);
         $message->setReceiptHandle($sqsMessage['ReceiptHandle']);
 
--- a/pkg/sqs/SqsConsumer.php
+++ b/pkg/sqs/SqsConsumer.php
@@ -207,6 +207,8 @@ class SqsConsumer implements Consumer
             $message->setProperties($headers[1]);
         }
 
+        $message->setMessageId($sqsMessage['MessageId']);
+
         return $message;
     }
 }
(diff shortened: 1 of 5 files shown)


pkg/sqs/SqsConsumer.php
```diff
@@ -185,6 +185,10 @@ protected function convertMessage(array $sqsMessage): SqsMessage
     {
         $message = $this->context->createMessage();
 
+        if (isset($sqsMessage['MessageId'])) {
```
makasim: No need for if. As far as I can see the field is there all the time.

BenoitLeveque: @makasim ok, i've update my code to remove the isset check and have also move the setMessage call to the end of the function, because the setHeader is overriding the message_id header value

BenoitLeveque:
@makasim any news about this pull request ?

pkg/sqs/SqsConsumer.php
makasim: The `if` is still there, isn't it? Should be removed as the `MessageId` is always there.

BenoitLeveque: oups, my bad i was sure to have pushed the deletion on this `if`

pkg/sqs/Tests/SqsConsumerTest.php
```diff
@@ -337,7 +338,7 @@ public function testShouldReceiveMessage()
 
         $this->assertInstanceOf(SqsMessage::class, $result);
         $this->assertEquals('The Body', $result->getBody());
-        $this->assertEquals(['hkey' => 'hvalue'], $result->getHeaders());
+        $this->assertEquals(['hkey' => 'hvalue', 'message_id' => 'theMessageId'], $result->getHeaders());
```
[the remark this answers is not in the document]
BenoitLeveque: If I change this, we will need to change this in all the other message type as we are reusing the test between services

Issue #1007, opened 2020-01-08
SQS - No AWS profile support


bgaillard:
For now the SQS transport does not allow to specify an AWS profile in the DSN string. This is problematic because using AWS profile is very common on developer machines.

We should be able to provide DSN like this one. 

```
sqs://aws?profile=staging&region=eu-west-1&version=latest&lazy=true
```

As a result we should let the AWS PHP SDK do its job to read the correct AWS credentials from the `~/.aws/credentials` file. The AWS PHP SDK will then used its INI file code to parse this file (see https://github.com/aws/aws-sdk-php/blob/master/src/Credentials/CredentialProvider.php#L391).

I'll provide a PR for this just after.

bgaillard:
@makasim thanks for this fast merge ! 

When do you think it will be possible to have this inside a composer release ?

PR #1008, opened 2020-01-08
SQS Transport - Add support for AWS profiles.

bgaillard:
This PR add support for AWS profiles, it reponds to issue https://github.com/php-enqueue/enqueue-dev/issues/1007.

Closes: #1007

### DIFF
--- a/pkg/sqs/SqsConnectionFactory.php
+++ b/pkg/sqs/SqsConnectionFactory.php
@@ -31,7 +31,8 @@ class SqsConnectionFactory implements ConnectionFactory
      *   'retries' => 3,              (int, default=int(3)) Configures the maximum number of allowed retries for a client (pass 0 to disable retries).
      *   'version' => '2012-11-05',   (string, required) The version of the webservice to utilize
      *   'lazy' => true,              Enable lazy connection (boolean)
-     *   'endpoint' => null           (string, default=null) The full URI of the webservice. This is only required when connecting to a custom endpoint e.g. localstack
+     *   'endpoint' => null,          (string, default=null) The full URI of the webservice. This is only required when connecting to a custom endpoint e.g. localstack
+     *   'profile' => null,           (string, default=null) The name of an AWS profile to used, if provided the SDK will attempt to read associated credentials from the ~/.aws/credentials file.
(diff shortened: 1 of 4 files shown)


Steveb-p:
@bgaillard you can use github-specific keywords to automatically close some issues when PRs are resolved - see https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

pkg/sqs/SqsConnectionFactory.php
```diff
@@ -120,10 +125,7 @@ private function parseDsn(string $dsn): array
         $dsn = Dsn::parseFirst($dsn);
 
         if ('sqs' !== $dsn->getSchemeProtocol()) {
-            throw new \LogicException(sprintf(
-                'The given scheme protocol "%s" is not supported. It must be "sqs"',
-                $dsn->getSchemeProtocol()
-            ));
+            throw new \LogicException(sprintf('The given scheme protocol "%s" is not supported. It must be "sqs"', $dsn->getSchemeProtocol()));
```
Steveb-p: This is only a code style change, and it violates the soft limit for characters in a single row (80-120, usually, see PSR https://www.php-fig.org/psr/psr-2/#23-lines).

bgaillard: Hi and thanks for your fast feedback :+1: . No problem i'll revert this in few minutes. 

Perhaps their is a missconfiguration with what's described in the https://github.com/php-enqueue/enqueue-dev/blob/master/docs/contribution.md guide because those lines were automatically updated by php-cs fixer.

Steveb-p: If it's true then keep it, otherwise CI will complain when testing. We will need to update php-cs ruleset then.

bgaillard: Ok, you're right I tested it with the change on one line and it make the CI fail. So I revert it to 3 lines.

bgaillard:
Ok, did not know this, thanks. 

Since your last review I just changed file https://github.com/php-enqueue/enqueue-dev/pull/1008/files#diff-663c8e41658b32f3c5e6dfd4bee59e96 to make CI pass.

I think everything is good to merge this now.

PR #1009, opened 2020-01-10
Catch consume error "Local: Broker transport failure" and continue consume

rdotter:
Catch the Kafka `Local: Broker transport failure` (error -195) error when consuming, this happens sometimes when consuming messages. When we restart the consumer everything is just OK.

Settings for connection with Kafka:
```
enqueue:
    kafka:
        transport:
            global:
                group.id: XXX
                metadata.broker.list: XXX
                enable.auto.offset.store: 'false'
                socket.keepalive.enable: 'true'
                topic.metadata.refresh.interval.ms: '30000'
```
@makasim


Steveb-p:
> RD_KAFKA_RESP_ERR__TRANSPORT - failed to connect to single broker, or connection error for single broker

So for now I agree we can allow silent discarding of this. :+1:

PR #1013, opened 2020-01-26
[rdkafka]S tore TopicConf in Context to prevent usage of deprecated Conf method

Steveb-p:


nick-zh:
@Steveb-p could you maybe resolve the conflict? Otherwise this looks good to me :v: 
I still think, we should push users to use `librdkafka:1.0` and up (since there were a looot of fixes),
but this a very good approach that doesn't rely on that.

nick-zh:
@Steveb-p can be closed now, was fixed in #1101

Steveb-p:
Thanks for the heads up @nick-zh :D

PR #1015, opened 2020-01-27
fix: allow absolute paths for sqlite transport

### LINKED ISSUE #780: Enqueue DBAL DSN invalid
Hi,

I am trying to setup the DBAL transport with SQLite in Symfony.
My enqueue.yml is:
```
enqueue:
    default:
        transport: '%env(resolve:DATABASE_URL)%'
        client: ~
```
An in my .env I have:
```
DATABASE_URL=sqlite:///%kernel.project_dir%/var/data.db
```

If I try to start Symfony console I get `Failed to parse DSN "sqlite:////*******/var/data.db"`

So I tried to tweak the DSN, but the only way to make it work is by putting a single /, but then if I try the command enqueue:setup-broker, I get `Failed to parse DSN "sqlite:////media/perso/PhpstormProjects/downloader/var/data.db"`

Am I doing something wrong?

cawolf:
Hi,

when switching from `amqp+lib` to `sqlite3` transport, we were not able to setup DSNs with absolute paths. It seems that `parse_url()` only allows more than 2 slashes for `file:` scheme and fails for other schemes (see [Notes in the documentation](https://www.php.net/manual/en/function.parse-url.php)). SQLite however happily uses [3 and more slashes for DSNs](https://www.doctrine-project.org/projects/doctrine-dbal/en/2.10/reference/configuration.html#connecting-using-a-url).

This PR proposes a special handling for `sqlite:` scheme and fixes #780. Relevant changes are lines 100-103, the rest is php-cs-fixer.

### DIFF
--- a/pkg/dbal/Tests/DbalConnectionFactoryTest.php
+++ b/pkg/dbal/Tests/DbalConnectionFactoryTest.php
@@ -28,4 +28,32 @@ class DbalConnectionFactoryTest extends TestCase
         $this->assertAttributeEquals(null, 'connection', $context);
         $this->assertAttributeInternalType('callable', 'connectionFactory', $context);
     }
+
+    public function testShouldParseGenericDSN()
+    {
+        $factory = new DbalConnectionFactory('pgsql+pdo://foo@bar');
+
+        $context = $factory->createContext();
+
+        $this->assertInstanceOf(DbalContext::class, $context);
+
+        $config = $context->getConfig();
+        $this->assertArrayHasKey('connection', $config);
+        $this->assertArrayHasKey('url', $config['connection']);
+        $this->assertEquals('pdo_pgsql://foo@bar', $config['connection']['url']);
+    }
+
+    public function testShouldParseSqliteAbsolutePathDSN()
+    {
+        $factory = new DbalConnectionFactory('sqlite+pdo:////tmp/some.sq3');
+
+        $context = $factory->createContext();
+
+        $this->assertInstanceOf(DbalContext::class, $context);
+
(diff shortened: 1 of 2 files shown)


makasim:
There is a better approach to the issue. Could you consider using Dsn class? Check the example https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/fs/FsConnectionFactory.php#L73

That would be great if you could add some tests too.

cawolf:
Oh great, I didn't check the other transport implementations, but this looks like a cleaner approach. I will update the PR accordingly.

pkg/dbal/Tests/DbalConnectionFactoryTest.php
```diff
@@ -28,4 +28,32 @@ public function testShouldCreateLazyContext()
         $this->assertAttributeEquals(null, 'connection', $context);
         $this->assertAttributeInternalType('callable', 'connectionFactory', $context);
     }
+
+    public function testShouldParseGenericDSN()
```
makasim: move tests to DbalConnectionFactoryConfigTest

cawolf: Are you sure? I'm not testing the config, I'm trying to verify that the factory creates a valid configuration and no information is lost in the process.

makasim: okay

cawolf:
Do you have more ideas for improving this PR? Or can it be merged?

makasim:
All is good.

Issue #1017, opened 2020-01-29
[GPS] GpsProducer unable to send to a queue


biforb4:
Hi, I'm using Enqueue 0.10 on Symfony 4.4 with symfony bundle. Trying to use the google pub sub transport, Unfortunately consuming messages doesn't seem to work.

Running
> /bin/console enqueue:consume --setup-broker -vvv default

results in
> In InvalidDestinationException.php line 17:
>[Interop\Queue\Exception\InvalidDestinationException]                                      
>The destination must be an instance of Enqueue\Gps\GpsTopic but got Enqueue\Gps\GpsQueue. 

Looking at the GpsProducer indeed it can only handle an instance of the GpsTopic.

I've compared this with AmqpProducer and it can send messages to both - queues and topics. Consuming messages with amqp transport works fine.

Is there a reason for this? Seems like using a amqp transport is impossible at the moment.

makasim:
I am not much into GPS logic. It might be a missing feature. Feel free to send a PR.

PR #1020, opened 2020-02-10
#1017 Allow GpsProducer to send messages to the queues

biforb4:


pkg/gps/GpsProducer.php
```diff
@@ -27,16 +27,20 @@ public function __construct(GpsContext $context)
...
+        $destination instanceof \Interop\Queue\Topic
+            ? InvalidDestinationException::assertDestinationInstanceOf($destination, GpsTopic::class)
+            : InvalidDestinationException::assertDestinationInstanceOf($destination, GpsQueue::class)
+        ;
         InvalidMessageException::assertMessageInstanceOf($message, GpsMessage::class);
 
+        $gpsTopicName = $destination instanceof GpsTopic ? $destination->getTopicName() : $destination->getQueueName();
         /** @var Topic $topic */
-        $topic = $this->context->getClient()->topic($destination->getTopicName());
+        $topic = $this->context->getClient()->topic($gpsTopicName);
```
makasim: Sorry, It does not make much sense to me. You use queue but publish to the topic. Could you elaborate on this?

makasim: I am not much into gps. But I recall that it is not possible to publish to gps queue directly, hence it has not been implemented.

biforb4: As far as I understand gps doesn't have a concept of queues, just topics and subscriptions. Hence I'd treat topic as a queue here. The problem is when clients router processor sends a message it sends it to the queue.

makasim: I'd like to hear more opinions on this. Especially whether the topic and queue are the same in Google PubSub world or not. 

I am not an expert.

makasim: Some links to refs or docs that prove it is same stuff would be ok too

biforb4:
There’s just no such thing as a queue in the gps. You’d always publish messages to a topic. Here’s some more explanation: https://cloud.google.com/pubsub/docs/overview#concepts

pkg/gps/GpsProducer.php
biforb4: There’s just no such thing as a queue in the gps. You’d always publish messages to a topic. Here’s some more explanation: https://cloud.google.com/pubsub/docs/overview#concepts

biforb4: I've been digging into gps transport code to get better understanding of the problem. 
I can see that currently we treat the pubsub topic as enqueue topic and pubsub subscription as queue. And that would be fine if not for for the fact that there are cases where enqueue tries to send a message to the queue. 

As is it not possible to publish anything to the pubsub subscription I believe we could treat the enqueue queue in two ways:
- as a topic for publishing messages
- as a queue for procsessing messages
(as per this PR)

biforb4: bump

ivansrdic: @makasim , If I may join the discussion - the change @biforb4 did is correct. I've just spent the whole day digging through enqueue and his suggestion is correct. 
We are switching from AWS to GCloud and without the change, enqueue is basically useless for GPS.
I really hope you will reconsider this PR.

Issue #821, opened 2019-04-10
[Beanstalkd] Unable to use delay


slaur:
I'm trying to use enqueue/pheanstalk with the symfony bundle.
Messages without delay are processed without error.
If I send a message with delay, the `setDeliveryDelay` method in `PheanstalkProducer` will throw an exception.
![ex](https://user-images.githubusercontent.com/1848190/55877402-3263c000-5b9a-11e9-985a-bf33ed810bcf.png)
Was someone able to use delayed jobs with beanstalkd and enqueue ?

makasim:
The logic is not implemented, feel free to provide a pull request that adds it.

slaur:
Thank you, I'll have a look.

slaur:
Use `$message->setHeader('delay', $delay);` instead of `$message->setDelay($delay);`

PR #1033, opened 2020-03-19
Implement DeliveryDelay, Priority and TimeToLive in PheanstalkProducer

likeuntomurphy:
I could not take advantage of the `MultiplierRetryStrategy` built in to the Symfony Messenger component, because the Pheanstalk producer did not fully implement the `setDeliveryDelay` method (see also #821). Even though I only needed the delay, I went ahead and fully implemented the priority and time to live functionality of the `Producer` interface as well.

The code changes in the PR essentially do three things:

1.  ensures that “delivery delay” and “time to live” are set as milliseconds on the producer but passed to the transport as seconds; 
2. ensures that the respective message values are used when priority, delay or TTL is null; and
3. ensures that any producer values do not persist for more than one send.

Each of these points has test coverage.

### DIFF
--- a/pkg/pheanstalk/PheanstalkProducer.php
+++ b/pkg/pheanstalk/PheanstalkProducer.php
@@ -52,11 +52,21 @@ class PheanstalkProducer implements Producer
             throw new \InvalidArgumentException(sprintf('Could not encode value into json. Error %s and message %s', json_last_error(), json_last_error_msg()));
         }
 
+        if (null !== $this->priority && null === $message->getHeader('priority')) {
+            $message->setPriority($this->priority);
+        }
+        if (null !== $this->deliveryDelay && null === $message->getHeader('delay')) {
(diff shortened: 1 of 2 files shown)


makasim:
Thank you for the contribution @likeuntomurphy. 

> ensures that any producer values do not persist for more than one send.

That's not true. In fact, the producer should use the values for all messages that go through it. That's home JMS defines it and other Enqueue transports work.

likeuntomurphy:
Thanks for the feedback, @makasim.

I realized that the enqueue transport will set the values on the producer for each send as long as the values are configured as options (https://github.com/sroze/messenger-enqueue-transport/blob/master/QueueInteropTransport.php#L140). So I removed the code that was resetting the values after each send.

And I was also wrong about how to prioritize the values. I copied a block of code from the Mongo producer and modified it for Beanstalk, and I updated the tests accordingly.

Issue #961, opened 2019-10-08
[Symfony bundle] Unable to test message queue


balabis:
According to the documentation, to test if a message has been sent - you have to have a TraceableProducer class. To get this class one should call `...client->getContainer()->get(TraceableProducer::class);`. Unfortunately, this yields an error: 
```
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException : You have requested a non-existent service "Enqueue\Client\TraceableProducer".
```
Would someone please explain to me how should I test that the message has been sent properly and why am I getting this error?

Steveb-p:
https://github.com/php-enqueue/enqueue-dev/blob/e02b63da7a5b2a86ba26e3b6f478af83827d6cec/pkg/enqueue/Symfony/Client/DependencyInjection/ClientFactory.php#L186-L191

...declares that `TraceableProducer` is only created if `traceable_producer` configuration parameter is set to `true` (which defaults to your current `debug` setting in Symfony).

Dump your configuration using `debug:config enqueue` command. If it is `true`, then see if maybe service defaulted to being `private`, therefore not available in `container`. If it is private, then your can use `ServiceSubscriberInterface` (Symfony concept, not related to enqueue) to access it - or regular dependency injection, especially if you're using Symfony's autowiring feature.

Steveb-p:
> Yeah, it shows that the service is not public. Thanks for the advice.
> What would be the appropriate way to test that the message has been delivered to the broker (in this case it's Redis)?

@balabis You actually don't have to. See:

https://github.com/php-enqueue/enqueue-dev/blob/97e2f99ca1533dbe4819db9206ee69301968af1b/pkg/redis/RedisProducer.php#L44-L72

Redis producer will throw an exception if connection Redis instance is not possible or Redis is unable to store it (if you're not familiar with Redis, keep in mind that saving to disk requires it to temporarily double it's RAM usage, which is a known pitfall).

If you haven't got an exception, then a message should be stored on Redis. You can check this by taking your Redis offline and trying to send.

Steveb-p:
@balabis You should not need to use another connection. It will also complicate your application unnecessarily and make it slower.

Instead, we could react to the fact that your connection is set with `auto_commit: false` and trigger commit ourselves. You could look into dbal package internals and see what changes need to be made to work with it.

[...]

PR #1042, opened 2020-04-06
Added: ability to choose different entity manager 

balabis:
Here the discussion: https://github.com/php-enqueue/enqueue-dev/issues/961#issuecomment-541027049

### DIFF
--- a/pkg/job-queue/Doctrine/JobStorage.php
+++ b/pkg/job-queue/Doctrine/JobStorage.php
@@ -46,14 +46,12 @@ class JobStorage
     /**
      * @param string $entityClass
      * @param string $uniqueTableName
-     * @param mixed  $entityManagerName
      */
-    public function __construct(ManagerRegistry $doctrine, $entityClass, $uniqueTableName, $entityManagerName = null)
+    public function __construct(ManagerRegistry $doctrine, $entityClass, $uniqueTableName)
     {
         $this->doctrine = $doctrine;
         $this->entityClass = $entityClass;
         $this->uniqueTableName = $uniqueTableName;
-        $this->entityManagerName = $entityManagerName;
     }
 
     /**
@@ -204,9 +202,7 @@ class JobStorage
     private function getEntityManager()
     {
         if (!$this->em) {
-            $this->em = empty($this->entityManagerName)
-                ? $this->doctrine->getManagerForClass($this->entityClass)
-                : $this->doctrine->getManager($this->entityManagerName);
+            $this->em = $this->doctrine->getManagerForClass($this->entityClass);
         }
         if (!$this->em->isOpen()) {
             $this->em = $this->doctrine->resetManager();
--- a/pkg/job-queue/Doctrine/JobStorage.php
+++ b/pkg/job-queue/Doctrine/JobStorage.php
@@ -46,9 +46,9 @@ class JobStorage
     /**
      * @param string $entityClass
      * @param string $uniqueTableName
-     * @param $entityManagerName
+     * @param mixed  $entityManagerName
      */
-    public function __construct(ManagerRegistry $doctrine, $entityClass, $uniqueTableName, $entityManagerName)
+    public function __construct(ManagerRegistry $doctrine, $entityClass, $uniqueTableName, $entityManagerName = null)
     {
         $this->doctrine = $doctrine;
         $this->entityClass = $entityClass;
--- a/pkg/job-queue/Doctrine/JobStorage.php
+++ b/pkg/job-queue/Doctrine/JobStorage.php
@@ -44,7 +44,6 @@ class JobStorage
     private $entityManagerName;
 
     /**
-     * @param ManagerRegistry $doctrine
      * @param string $entityClass
      * @param string $uniqueTableName
      * @param $entityManagerName
@@ -97,7 +96,6 @@ class JobStorage
 
     /**
      * @param string $name
-     * @param Job    $rootJob
      *
      * @return Job
      */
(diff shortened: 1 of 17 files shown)


balabis:
Hey, @Steveb-p, would you mind checking this PR?

Steveb-p:
I'll try to look it up more thoroughly in the evening.

Steveb-p:
@balabis Have you tested this against your application?

I've looked through the extension and `Enqueue\JobQueue\Tests\Functional\Job\JobStorageTest`, and wasn't able to find the part that injects enqueue's job entity mapping into Doctrine. Are you sure it's made automatically?

```yml
enqueue:
    default:
        # plus basic bundle configuration

        job: true
        
        # by default bundle will add a default orm mapping configuration 
        # if you define custom mappings, you can specify which entity manager to use here 
        entity_manager_name: ~
```
This part from the original documentation was responsible for injecting `Job` metadata into Doctrine entity manager. I'm pretty sure it's required, since I haven't found any other references to mapping metadata in bundle files.

I'm pretty sure tests would fail the moment you'd start using a different doctrine entity manager than default - reason being that `EntityManager` would not contain metadata required to work with `Job` entities.

To test your branch against your application you can use composer's built-in functionality which allows using one's fork instead of a library.

In your case it should look something like this:
```
{
    "repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/balabis/enqueue-dev.git"
        }
    ],
    "require": {
        "enqueue/job-queue": "dev-master"
    }
}
```

balabis:
`enqueue-bundle/DependencyInjection/EnqueueExtension` lines 179-198 prepends the Doctrine mappings by default:
```
foreach ($container->getExtensionConfig('doctrine') as $config) {
            // do not register mappings if dbal not configured.
            if (!empty($config['dbal'])) {
                $rc = new \ReflectionClass(Job::class);
                $jobQueueRootDir = dirname($rc->getFileName());
                $container->prependExtensionConfig('doctrine', [
                    'orm' => [
                        'mappings' => [
                            'enqueue_job_queue' => [
                                'is_bundle' => false,
                                'type' => 'xml',
                                'dir' => $jobQueueRootDir.'/Doctrine/mapping',
                                'prefix' => 'Enqueue\JobQueue\Doctrine\Entity',
                            ],
                        ],
                    ],
                ]);
                break;
            }
        }
```

The part I removed from docs suggests the user that he should define the Doctrine mappings manually, but in reality that mapping is injected behind the scenes with the code above. 

So this default mapping is always there, but the user, if he so choses, can define his own mappings (thats what we did). And now you can define as many mappings as you want, but JobStorage will always use the default one. 

I have used this code with our application. If I specify `entity_manager_name`, JobStorage will use it and queue will be consumed using this entity manager.

Steveb-p:
Well, I'm blind apparently. Thanks for finding it. Goes to show that I'm not really that much more familiar with enqueue-bundle itself than you :smile: 

However, instead of injecting an entity name into `JobStorage` and creating a mapping manually (which you still have to do, so that part in documentation about adding a mapping would be even more important :grey_exclamation: ), I would make `EnqueueExtension` inject the mapping properly to specified entity manager.
This would prevent class metadata from existing in multiple managers and would allow doctrine's `ManagerRegistry` to properly resolve manager when `getManagerByClass` is called (in user code, perhaps).

Could you please see if this is possible there?

balabis:
Ok, will see what I can do.

balabis:
What do you think if instead of having:
```
foreach ($container->getExtensionConfig('doctrine') as $config) {
            // do not register mappings if dbal not configured.
            if (!empty($config['dbal'])) {
                $rc = new \ReflectionClass(Job::class);
                $jobQueueRootDir = dirname($rc->getFileName());
                $container->prependExtensionConfig('doctrine', [
                    'orm' => [
                        'mappings' => [
                            'enqueue_job_queue' => [
                                'is_bundle' => false,
                                'type' => 'xml',
                                'dir' => $jobQueueRootDir.'/Doctrine/mapping',
                                'prefix' => 'Enqueue\JobQueue\Doctrine\Entity',
                            ],
                        ],
                    ],
                ]);
                break;
            }
        }
```
We'd have something like this:
```
foreach ($container->getExtensionConfig('doctrine') as $config) {
            // do not register mappings if dbal not configured.
            if (!empty($config['dbal'])) {
                $rc = new \ReflectionClass(Job::class);
                $jobQueueRootDir = dirname($rc->getFileName());
                $mappingExists = false;

                array_walk_recursive($config['orm'], function ($item, $key) use (&$mappingExists) {
                   if ($key === 'prefix' && $item === 'Enqueue\JobQueue\Doctrine\Entity') {
                       $mappingExists = true;
                   }
                });

                if (!$mappingExists) {
                    $container->prependExtensionConfig('doctrine', [
                        'orm' => [
                            'mappings' => [
                                'enqueue_job_queue' => [
                                    'is_bundle' => false,
                                    'type' => 'xml',
                                    'dir' => $jobQueueRootDir.'/Doctrine/mapping',
                                    'prefix' => 'Enqueue\JobQueue\Doctrine\Entity',
                                ],
                            ],
                        ],
                    ]);
                }

                break;
            }
        }
```
The idea here would be to prepend the default mapping only if the user has not defined any mappings.

Steveb-p:
> The idea here would be to prepend the default mapping only if the user has not defined any mappings.

I would consider a different approach. Instead of trying to establish if there are any mappings provided by the user, create a configuration setting to explicitly do so - with the default being `true`, to maintain current behavior (which would nicely prevent BC breaks).

This is also important due to the fact that you can technically provide your own Job classes as far as I'm aware (might be wrong though - as I said, I'm not familiar with JobQueue component). Code proposed would inject mapping in this case anyway.

balabis:
Hey, @Steveb-p, mind checking the PR again?

Steveb-p:
> Hey, @Steveb-p, mind checking the PR again?

Yeah, will do in the evening or tomorrow morning.

balabis:
Hey, @Steveb-p , have you checked it?

pkg/job-queue/Doctrine/JobStorage.php
```diff
@@ -39,9 +39,13 @@ class JobStorage
     private $uniqueTableName;
 
     /**
-     * @param ManagerRegistry $doctrine
-     * @param string          $entityClass
-     * @param string          $uniqueTableName
+     * @var string
+     */
+    private $entityManagerName;
```
Steveb-p, line 44: I think it's some leftover code from when there was still the concept of setting a particular entity manager?

Steveb-p (commented):
Sorry for taking so long.

There are two things I'd like you to look at at this point. Otherwise than that it look ok to me - I know that code style fixer might be a PITA that causes a lot of changes here (which can probably be avoided by squashing commits and amending/removing those changes) but I think personally it's ok to fix them right away.

balabis:
Thanks @Steveb-p 
I've edited these two things and created a new PR #1081 for it.

PR #1043, opened 2020-04-06
Added support for ramsey/uuid 4.0

a-menshchikov:


snapshotpl:
@makasim can you release this change?

makasim:
this PR updates only root composer.json but the version has to be updated in all packages.

levelfivehub:
Does this mean, it is best to remain with ramsey/uuid 3 until further notice?

atrauzzi:
Is there any chance we can get this fix released?  It's making installation very difficult...

PR #1052, opened 2020-04-20
Fix RedisProducer MessageId

### LINKED ISSUE #1051: RedisProducer not respecting custom MessageId property
Hi, I am trying to upgrade from 0.8 to 0.9. I've found an issue in the RedisProducer.

At [this line](https://github.com/php-enqueue/redis/blob/0.9.12/RedisProducer.php#L49) the message id is entirely replaced.
```php
    $message->setMessageId(Uuid::uuid4()->toString());
    $message->setHeader('attempts', 0);
```

Looking at the [Producer](https://github.com/php-enqueue/enqueue-dev/blob/0.9.15/pkg/enqueue/Client/Producer.php#L110-L112) class, the message id is only set if it doesn't exist.
```php
    if (!$message->getMessageId()) {
        $message->setMessageId(UUID::generate());
    }
```

In my code, I set the message id before I send the message. I use the message id property as the Job's owner id, but RedisProducer is not respecting my choice and my consumers are not working properly, because they are not able to find the job by that owner id.

I am using the Job integration a bit different, creating the jobs before sending the messages because my frontend client expects a "promise" resource.

In the meantime I am using another message property as a workaround, but I think the RedisProducer should check if the message id is empty in order to set a value.

Thanks!

ixarlie:
Fixes #1051

### DIFF
--- a/pkg/redis/RedisProducer.php
+++ b/pkg/redis/RedisProducer.php
@@ -29,9 +29,6 @@ class RedisProducer implements Producer
      */
     private $deliveryDelay;
 
-    /**
-     * @param RedisContext $context
-     */
     public function __construct(RedisContext $context)
     {
         $this->context = $context;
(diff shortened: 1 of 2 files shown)


ixarlie:
Hi @makasim , I am not sure why the tests are failing. please advise.

ixarlie:
Hi @makasim some news about this? Thanks!

PR #1055, opened 2020-05-04
Added: additional parameter for redelivery delay

balabis:
This feature would introduce some more flexibility manipulating redelivery delays. For example, you know some messages might be slower to process than usual, but it can happen rarely so you don't want to increase the redelivery time for all the retries. By having this parameter you could set an initial redelivery time to be later, but also keep the regular `redelivery_delay` lower so it wouldn't affect all the retries.


balabis:
Hey @Steveb-p, maybe you could check this PR when you'd have some time? :)

Steveb-p:
I don't really understand the purpose at first glance. I've skimmed through the code and I can't figure out what difference there is between `redelivery_delay` and `initial_delay`.

Name suggests that `initial_delay` postpones first message processing, but I don't really know what purpose would that serve in normal use-case. I'll have a second look after work.

PR #1070, opened 2020-06-22
docs: update kafka dsn

qkdreyer:
https://github.com/php-enqueue/rdkafka/commit/d91547e0e1195c521d859be617c9b1111a69ba17

### DIFF
--- a/docs/transport/kafka.md
+++ b/docs/transport/kafka.md
@@ -154,7 +154,7 @@ Set your enqueue to use rdkafka as your transport
 
 enqueue:
     default:
-        transport: "rdkafka:"
+        transport: "kafka:"
         client: ~
 ```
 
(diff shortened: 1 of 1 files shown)


Steveb-p:
I believe both still work @qkdreyer ?

qkdreyer:
Sadly not

https://github.com/php-enqueue/rdkafka/blob/f63b08d35647aab31f8731293718931e08628b57/RdKafkaConnectionFactory.php#L78-L80

Steveb-p:
I'm pretty sure it still does work when you put `rdkafka://`into Symfony configuration as `dsn` property specifically. My own application still relies on that and I'm on version 0.10.1.

The `parseDsn` method you linked only takes effect if configuration is a string, which in the example provided it is not - it is in fact an array.

If it were:
```yaml
enqueue:
    default:
        transport: "rdkafka://"
```
then it would throw an exception.

Steveb-p:
The `parseDsn` method that requires specifically `kafka://` string (to get schema === `kafka`) is only called in cases where configuration is a string, i.e.:
```yaml
enqueue:
    default:
        transport: "rdkafka:"
```

When array is used - as in second part of your PR - it will work correctly as schema is not checked in that case.

```yaml
enqueue:
    default:
        transport:
            dsn: "rdkafka://"
```

In principle I agree - we should either accept both "schemes" or single one in all circumstances. However I'm not sure if the transport selection will work correctly for the second case (array).

I'll look how this behaves in my own application later in the day.

PR #1080, opened 2020-08-13
feat(dbal): add the dropDataBaseTable() method to the context

fancyweb:
`DbalContext` has an helper method to create the database table. Having the reverse method to drop the database table would be helpful for DX when the table is created / dropped programmatically.


fancyweb:
@makasim Could you consider reopening this PR?

PR #1081, opened 2020-08-20
Added: ability to choose different entity manager

balabis:
Fixed issues of #1042

### DIFF
--- a/pkg/enqueue-bundle/DependencyInjection/Configuration.php
+++ b/pkg/enqueue-bundle/DependencyInjection/Configuration.php
@@ -95,7 +95,7 @@ final class Configuration implements ConfigurationInterface
 
         return (new ArrayNodeDefinition('job'))
             ->children()
-                ->scalarNode('entity_manager_name')->defaultNull()->end()
+                ->booleanNode('default_mapping')->defaultTrue()->end()
             ->end()
             ->addDefaultsIfNotSet()
             ->canBeEnabled()
--- a/pkg/enqueue-bundle/DependencyInjection/Configuration.php
+++ b/pkg/enqueue-bundle/DependencyInjection/Configuration.php
@@ -95,7 +95,10 @@ final class Configuration implements ConfigurationInterface
 
         return (new ArrayNodeDefinition('job'))
             ->children()
-                ->booleanNode('default_mapping')->defaultTrue()->end()
+                ->booleanNode('default_mapping')
+                    ->defaultTrue()
+                    ->info('Adds bundle\'s default Job entity mapping to application\'s entity manager')
+                ->end()
             ->end()
             ->addDefaultsIfNotSet()
             ->canBeEnabled()
--- a/pkg/enqueue-bundle/DependencyInjection/Configuration.php
+++ b/pkg/enqueue-bundle/DependencyInjection/Configuration.php
@@ -94,6 +94,9 @@ final class Configuration implements ConfigurationInterface
         }
 
         return (new ArrayNodeDefinition('job'))
+            ->children()
+                ->scalarNode('entity_manager_name')->defaultNull()->end()
+            ->end()
             ->addDefaultsIfNotSet()
             ->canBeEnabled()
         ;
--- a/pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php
+++ b/pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php
@@ -98,11 +98,6 @@ final class EnqueueExtension extends Extension implements PrependExtensionInterf
                     throw new \LogicException('Job-queue supports only default configuration.');
                 }
 
-                $container->setParameter(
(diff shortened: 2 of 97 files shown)


pkg/enqueue-bundle/DependencyInjection/Configuration.php
```diff
@@ -94,6 +94,9 @@ private function getJobConfiguration(): ArrayNodeDefinition
         }
 
         return (new ArrayNodeDefinition('job'))
+            ->children()
+                ->booleanNode('default_mapping')->defaultTrue()->end()
+            ->end()
```
Steveb-p, line 99: Actually we can use this place to add information about what this property does:

```diff
            ->children()
-                ->booleanNode('default_mapping')->defaultTrue()->end()
+                ->booleanNode('default_mapping')
+                    ->defaultTrue()
+                    ->info('Adds bundle's default Job entity mapping to application's entity manager')
+                ->end()
            ->end()
```

Steveb-p:
LGTM. All that is left is for @makasim to review.

Steveb-p (approved):

pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php
```diff
@@ -171,6 +171,14 @@ private function registerJobQueueDoctrineEntityMapping(ContainerBuilder $contain
             return;
         }
 
+        foreach ($container->getExtensionConfig('enqueue') as $modules) {
+            foreach ($modules as $config) {
+                if (isset($config['job']) && false === $config['job']['default_mapping']) {
```
makasim, line 176: That does not look accurate. For example one config defines `false` and the next defines `true`. The merged value might be `true` but this logic would not load default entites.

There must be a method that allows getting merged version of the config. But I am not sure. Could you please look for it?

makasim (changes requested):

pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php
balabis: Sorry for taking so long.
I was not able to find a solution for this yet, but this part of code is happening before the merging of configuration, no?

makasim: You should merge it manually and use the result in condition.

makasim: Symfony provides tools for that.

makasim: I believe

Steveb-p: https://symfony.com/doc/current/bundles/configuration.html#processing-the-configs-array

balabis:
Hey @Steveb-p , maybe you know what exactly causes the tests to fail?

`TypeError: Argument 1 passed to Enqueue\Doctrine\DoctrineConnectionFactoryFactory::__construct() must be an instance of Doctrine\Common\Persistence\ManagerRegistry, instance of Double\stdClass\P1 given`

Before the last code fix, all tests passed, but the `DoctrineConnectionFactoryFactory` constructor argument was the same as it is now, though locally, I can see that the 1st argument passed to the constructor is `Double\ManagerRegistry\P1`

balabis:
hey @makasim, would you mind reviewing this PR one more time?

pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php
```diff
@@ -171,6 +171,18 @@ private function registerJobQueueDoctrineEntityMapping(ContainerBuilder $contain
             return;
         }
 
+        $config = $container->getExtensionConfig('enqueue');
+
+        if (!empty($config)) {
+            $processedConfig = $this->processConfiguration(new Configuration(false), $config);
+
+            foreach ($processedConfig as $name => $modules) {
+                if (isset($modules['job']) && false === $modules['job']['default_mapping']) {
```
makasim, line 180: I think we can avoid `for` loop here. 

Something like this should do:

```php
isset($processedConfig['enqueue']['job']['default_mapping']) && $processedConfig['enqueue']['job']['default_mapping'] == false
```

makasim (changes requested):

pkg/enqueue-bundle/DependencyInjection/EnqueueExtension.php
balabis: Hm, `$processedConfig` does not have `enqueue` key. By default it would have a `default` key, but I guess it could be any string defined by the user, no? In the documentation it is defined as `key`:

Screenshot 2020-11-24 at 10 38 25

makasim: well, then it's ok.

PR #1087, opened 2020-09-01
Solves binding Headers Exchange with Queue using custom arguments

dgafka:
Solves binding `Queue` with `Headers Exchange` using custom arguments.

Screenshot from 2020-09-01 14-01-42

Screenshot from 2020-09-01 13-38-34


dgafka:
Not sure why tests are failing. 
Failing tests comes from different package, so it looks like master branch on different repository is failing.

@makasim Can you confirm?

PR #1093, opened 2020-09-28
Allow false-y values for unsupported options

atrauzzi:
This is to accommodate any situations where `0` is passed via configuration or even defaults from other layers (in my case, laravel).


makasim:
There is a code style issue. Could you please fix it?

makasim:
btw, master is green now

atrauzzi:
Updated!

atrauzzi:
Thank you!

Issue #1097, opened 2020-09-29
STOMP: "Was not possible to write frame!" causing zombie connections


atrauzzi:
I'm able to somewhat reliably reproduce a situation where my queue worker in Laravel ends up no longer able to successfully connect to a broker.  It takes a few consecutive jobs before it happens, but eventually my one queue gets the exception `"Was not possible to write frame!"`.

Thereafter, any attempt to send messages to the same broker end up reporting the same exception. It seems like [the connection becomes corrupted here](https://github.com/stomp-php/stomp-php/blob/master/src/Network/Connection.php#L487) and no recovery is attempted by enqueue or the stomp-php library itself.

Is there any general advice on how to handle this or is this something that enqueue can catch and recreate its connection?

atrauzzi:
@makasim - Can totally do a screen share with you on this one to get a better understanding if necessary. If the fix is simple enough, I'm always happy to contribute more. :smile:

makasim:
I suspect that the server closes connection due to lack of activity. Do the jobs arrive one after another or is there any pause between them ?

makasim:
There is a message to keep the connection alive https://github.com/stomp-php/stomp-php/blob/master/src/Network/Connection.php#L681

atrauzzi:
I believe the keep alive / heartbeating is working as I'm able to sustain a connection for prolonged periods of time.  Even overnight! :laughing: 

This situation seems to be intermittent. Almost like if I trigger a message, wait a bit and then trigger another, I get the error.  Around 8-10 seconds.

For fun, I tried this in `StompProducer` and enqueue is able to "recover":

```
try {
            $this->stomp->send($destination->getQueueName(), $stompMessage);
        }
        catch (ConnectionException $ex) {
        
            dump("IT BORKED!!!!");
        
            $this->stomp->disconnect(true);
            $this->stomp->connect();
        
            $this->stomp->send($destination->getQueueName(), $stompMessage);
        }
```

makasim:
You can add a consumption extension that calls [keepAlive](https://github.com/stomp-php/stomp-php/blob/master/src/Network/Connection.php#L681) method. Also, set lower receive timeout than connection timeout so the socket gives control back to PHP and consumption extensions are executed. That would do the trick. 

The solution does not solve all the possible cases. If you have a long-running job you might still run into the issue. More on it [here](https://github.com/php-enqueue/enqueue-dev/pull/658)

atrauzzi:
[...]

PR #1099, opened 2020-09-29
Fix - Add automatic reconnect support for STOMP producers

atrauzzi:
Closes: https://github.com/php-enqueue/enqueue-dev/issues/1097

---

To detail my scenario, I'm running a long-lived PHP process (laravel queue worker) that **listens on one connection**, but **can occasionally publish to another connection**. The connection used for _listening_ doesn't suffer from any issues because enqueue successfully sets up heartbeats, anticipating the fact that consumers tend to be long lived. For producers however, there might be the assumption that they are transient (web requests) and as such, don't require connections to live any longer than it takes for the request thread to finish all its work.

---

@makasim - My apologies if I've missed something or if this seems like a shoehorn fix / neglects any other concerns. After doing some digging around however I am starting to wonder if this might actually be a desirable approach.

What I've noticed in the `stomp-php` library is that their implementation of `Client::isConnected` has the potential to report false positives for the state of the connection and that without getting into a more convoluted fix, can only be "tested" by attempting to use it.

With the code fix in this PR, failures to write are jogged with one last-ditch attempt to connect. My long lived mixed _consumer+producer_ processes are able to send messages and reestablish connections on-demand with no ceremony or boilerplate.

I'd like to add the caveat that this is not my preferred strategy and would favour something more explicit, but just based on the fact that `StompProducer` isn't set up for configuration and the aforementioned false-positives from `stomp-php`, this may do well for all scenarios.


pkg/stomp/StompProducer.php
```diff
@@ -39,7 +40,18 @@ public function send(Destination $destination, Message $message): void
 
         $stompMessage = new StompLibMessage($message->getBody(), $headers);
 
-        $this->stomp->send($destination->getQueueName(), $stompMessage);
+        try {
+            $this->stomp->send($destination->getQueueName(), $stompMessage);
+        } catch (ConnectionException $ex) {
+            if (!$this->stomp->isConnected()) {
```
makasim, line 46: what this `if` for?

makasim:
>Do we not want this?

Yes, indeed. You should not commit lock file in libraries in contrast to application.

pkg/stomp/StompProducer.php
atrauzzi: I had originally placed it there so that if something explicitly closed the connection, it would be respected (rather than accidentally forcing it back on).  Basically to give _some_ hope for integrity of the state when it's `false`.

(It's just `true` that can't be trusted.)

atrauzzi:
> Yes, indeed. You should not commit lock file in libraries in contrast to application.

Hmm, I always lock my dependencies regardless of project type, but no matter! :smile: 

---

I'm going through some additional testing right now and I'm noticing one other issue with `php-stomp` where it claims to have successfully written, but actually hasn't.  I end up having to pump quite a few messages through before it reports a failed write.

I'll spend some more time today testing and will report back.

atrauzzi:
@makasim - So after looking a little deeper into it, it looks like there's another issue with `stomp-php` when sending messages where it occasionally fails to detect a dead connection.  I am able to reproduce this consistently.

The sequence looks roughly like:

 - Run a broker that connects to `queue-1` and listens.
 - Worker receives a message on `queue-1` and goes to handle it.
 - Worker requires a message to be written to `queue-2`
 - Worker connects to `queue-2` and write; the message is successful. This leaves the connection open with no heartbeat.
   > note: Up until this point, I think things are fine. If `queue-1` works steadily for a while, messages sent from the worker to `queue-2` are safe. This steady work simulates a heartbeat and prevents any dead connection detection on the broker from triggering.
 - If after a while (connection timeout window), a new message arrives on `queue-1` and the worker goes to process it similar to before, it ends up unable to deliver any message to `queue-2`.  The connection to the broker for `queue-2` is effectively broken, but internally to `stomp-php`, it thinks it's still good. Even to the point of attempting to write bytes without triggering an error.

---

Originally I thought `stomp-php` was simply not heartbeating a connection and not detecting dead connections, but now I'm able to see it also try to use its connection resource without throwing the `"Was not possible to write frame!"` exception. :scream: 

I've added another improvement to the PR which we can discuss further, it unfortunately has the impact of requiring one connection per message.  Again though, I'm not sure how to workaround this and in fairness, I can see how this hasn't been observed until now because I could see the majority of projects using this library only dispatching from short-lived web requests.

makasim:
To find out that connection is closed ( by the server) you have to read from the socket. You will get EOF or timeout. 

There is no way to detect a closed connection by writing to it. Timeout is the only option.

Is there a way to set write timeout ? That could be used as a sign of closed connection and trigger reconnect.

atrauzzi:
Oh thanks!  I'll tinker around with your suggestion and reply here with what I figure out.  Definitely need to be able to come up with a way to gracefully handle this.

I'll see how long I have to go before something has to happen upstream :laughing:

atrauzzi:
@makasim - After spending a fair bit of time on it, I've been unable to come up with a way to jog or test the connection.  Can you think of any concrete solutions that might get around this issue?

From my vantage point, I see only a few options:

 - Have enqueue send heartbeats for non-queue producer connections
 - Somehow figure out a way to jog the connection prior to using it
 - Allow enqueue to treat connections that don't have a heartbeat to be considered "transient" and destroy them after each message sent.

IMO, I feel like the 3rd option is the most "predictable" philosophy and also aligns message publishing with how it's functioning right now during web requests.

atrauzzi:
Just a note in case you're at things early today, will be pushing up a new approach that I think I have working really nicely here...

pkg/stomp/StompContext.php
```diff
@@ -38,10 +38,16 @@ class StompContext implements Context
...
+     * @var bool
+     */
+    private $transient;
+
     /**
      * @param BufferedStompClient|callable $stomp
+     * @param mixed                        $transient
      */
-    public function __construct($stomp, string $extensionType)
+    public function __construct($stomp, string $extensionType, $transient = true)
```
atrauzzi, line 50: Connections start out as "transient" which means that they will use disconnect behaviour.  I have it here in the constructor as configurable if for whatever reason a connection needs to start out long-lived.

pkg/stomp/StompContext.php
```diff
@@ -173,6 +180,8 @@ public function createConsumer(Destination $destination): Consumer
     {
         InvalidDestinationException::assertDestinationInstanceOf($destination, StompDestination::class);
 
+        $this->transient = false;
```
atrauzzi, line 183: As soon as the first consumer is created, we go long-lived and disable all transient functionality so that queues with heartbeats don't have their connections pulled out from under them.

pkg/stomp/StompContext.php
```diff
@@ -181,6 +190,10 @@ public function createConsumer(Destination $destination): Consumer
      */
     public function createProducer(): Producer
     {
+        if ($this->transient && true == $this->stomp) {
+            $this->stomp->disconnect();
+        }
```
atrauzzi, line 195: Simply triggering a disconnect is enough, as underlying layers can rely on their `false` state around connections, things pick up without any additional fuss at the enqueue level.

pkg/stomp/StompContext.php
```diff
@@ -202,14 +215,20 @@ public function purgeQueue(Queue $queue): void
...
+    private function createStomp(): BufferedStompClient
+    {
+        $stomp = call_user_func($this->stompFactory);
+
+        if (false == $stomp instanceof BufferedStompClient) {
+            throw new \LogicException(sprintf('The factory must return instance of BufferedStompClient. It returns %s', is_object($stomp) ? get_class($stomp) : gettype($stomp)));
+        }
+
+        return $stomp;
+    }
```
atrauzzi, line 233: Hope you don't mind, I broke this out. :smile:

atrauzzi:
@makasim - Updates made and I'm really happy with this, it's working great and has greatly improved the stability and flexibility of enqueue STOMP! :smiley:

makasim:
To clarify:

You have to connections: one for consumers and another for publisher. The problem happens when you want to publish a message from a consumer ?

makasim:
For example https://github.com/php-enqueue/enqueue-dev/blob/0bbba640e796bcf56066900718580a0ac384df58/pkg/enqueue-bundle/Consumption/Extension/DoctrinePingConnectionExtension.php

atrauzzi:
Hmm, let's see if I can do this justice :laughing: 

I think part of the issue is around the current division of responsibilities.  We need to do the detection at the `Context` because `StompConsumer` and `StompProducer` don't know if they should start out transient as well as when or whether they should stop being transient at some point.

 - `StompContext` only has the authority to manage connections, but never uses them
 - `StompProducer` and `StompConsumer` only have the authority to use connections , but not manage them

For the current design:

 - `StompContext` has responsibility for both `createProducer` and `createConsumer`
 - `StompContext` is responsible for triggering the creation and holding references to connections via `StompContext::createStomp`, this makes `StompContext` the first point of integration with `stomp-php`
 - So long as a `StompContext` is only ever used for _producers_, then it's better to preserve the one message per connection semantics by bouncing the connection before any attempt to use it.
   - This is because if a `StompContext` only has `createProducer` called, it almost definitely doesn't have a heartbeat configured
   - `createProducer` is called multiple times to send messages.  `StompProducers` are not reused and do not have sole authority over a connection, so it made sense to me to reset the connection at the level where it's managed.

In theory I _could_ add a `$transient` parameter to `StompProducer` instances, but it seems somewhat out of place as their current design is focused around using the connection and not managing it (as noted before).

Apologies, I'm hopefully explaining this correctly and in such a way that does the current design justice :laughing:

makasim:
I am not that much into Laravel specifics. In general the idea is next

* you've got a message
* forced disconnect-connect of publisher connection, like `$conn->reconnect();`
* consume the received message. 

Once you get a new message do the same.

atrauzzi:
That's for receiving messages, which is not a problem here.

The issue I'm observing is when I try to send to a queue that _is not_ being listened on, from a long-lived process.

atrauzzi:
@makasim - Is there anything specific I can do to help get this merged?  At this point my testing locally has proven the fix in this MR to be quite robust for long lived processes. :smile:

atrauzzi:
Hey @makasim, just checking in again.  I'd love to get this merged so that we'll be able to use STOMP from PHP in some upcoming work.  Everything is still working very smoothly with my local version of this fix.

makasim:
Honestly, I am not completely  sold on this solution. Let's discuss the issue over a call. What about tomorrow?

atrauzzi:
Absolutely, I can definitely do that, I'm in the CDT timezone, happy to accommodate.

It's worth noting that the transient flag is only in effect when the connection is going to be short-lived anyway. So there's very little risk of this breaking consumers. It's just causing the connection to recycle prior to use.

atrauzzi:
@makasim - If you're still available and unsure about this, I followed you on twitter if you wanna DM any details for a quick screen share or chat.

Again, I've been running this patch locally from two different queue workers that interact over the broker since figuring it out.  Have made sure to come back and test dispatches.  Consumers and heartbeats haven't been negatively impacted. My scenario also tests consuming and dispatching from the same queue, so the `transient` flag is being managed correctly.

makasim:
@atrauzzi  Sent you a message

pkg/stomp/StompConnectionFactory.php
```diff
@@ -74,14 +74,14 @@ public function __construct($config = 'stomp:')
...
     {
-        if ($this->config['lazy']) {
-            return new StompContext(
-                function () { return $this->establishConnection(); },
-                $this->config['target']
-            );
-        }
+        $stomp = $this->config['lazy']
+            ? function () { return $this->establishConnection(); }
+        : $this->establishConnection();
```
atrauzzi, line 79: Normally I indent lines like `79`, but the code style suggested this, apologies!

pkg/stomp/StompContext.php
```diff
@@ -38,10 +38,15 @@ class StompContext implements Context
...
+    /**
+     * @var bool
+     */
+    private $transient;
+
     /**
      * @param BufferedStompClient|callable $stomp
      */
-    public function __construct($stomp, string $extensionType)
+    public function __construct($stomp, string $extensionType, bool $detectTransientConnections = false)
```
atrauzzi, line 49: Here is where we default to "off".

atrauzzi:
Good morning @makasim! :laughing: 

As per our chat earlier today, I've wired this up to the configuration obtained in `StompConnectionFactory`. Pretty straightforward I'd say and preserves all old behaviour by default.

makasim:
@atrauzzi some tests are failing

atrauzzi:
@makasim - Okay, I think the two that are failing now are okay?

makasim (approved):

atrauzzi:
Wonderful, thank you @makasim.  Will this be automatically released?

makasim:
Semi-automatically... 

I'll do it.

makasim:
Thank you too!

atrauzzi:
Hey @makasim - did the release end up going through?  I noticed that I'm still seeing version `0.10.4` as the latest. :smile:

makasim:
done https://github.com/php-enqueue/stomp/releases/tag/0.10.5

PR #1101, opened 2020-10-05
[rdkafka] remove topic conf, deprecated

### LINKED ISSUE #970: [rdkafka] RdKafka\Conf::setDefaultTopicConf() is deprecated
PHP rd-kafka 4.0 throwing the following exception:

```
In RdKafkaContext.php line 193:
                                                              
  Function RdKafka\Conf::setDefaultTopicConf() is deprecated  
```

https://github.com/php-enqueue/rdkafka/blob/master/RdKafkaContext.php#L218

nick-zh:
This will fix #970, the use of topic conf is deprecated, all configs should be set over the global config.
To avoid a break / major, i adapted this internally where config on both levels (global, topic) will be treated / set equally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ install:
     - echo "memory_limit=2048M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
     - php ./bin/fix-symfony-version.php "$SYMFONY_VERSION"
     - composer install
-    - ./bin/install-librdkafka.sh
+    - sed -i 's/525568/16777471/' vendor/kwn/php-rdkafka-stubs/stubs/constants.php
     - if [ "$PREPARE_CONTAINER" = true ]; then docker --version; fi
     - if [ "$PREPARE_CONTAINER" = true ]; then docker-compose --version; fi
     - if [ "$PREPARE_CONTAINER" = true ]; then bin/dev -b; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,7 @@ install:
     - echo "memory_limit=2048M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
     - php ./bin/fix-symfony-version.php "$SYMFONY_VERSION"
     - composer install
+    - ./bin/install-librdkafka.sh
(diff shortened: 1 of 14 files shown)


nick-zh:
So the failing test is due to the fact that you guys are using `librdkafka:0.11.1`, i created another PR to use the lowest version that `php-rdkafka` supports, but my PR will only work against `librdkafka:1.x` (which is also already 1.5y old)

makasim:
@nick-zh is it ready ?

nick-zh:
@makasim so if we go with my PR, it means dropping support for anything below `librdkafka:1.x`, if you still want to support `librdkafka:0.11.6` (last version before 1.x), we need to update #1013 from @Steveb-p, up to you ✌️

nick-zh:
If you want to go with my PR, maybe i could add an exception when somebody tries to use it with a lib version that is not supported, let me know

makasim:
yes, please

nick-zh:
will get right no it

nick-zh:
@makasim done, but i think you need to rebuild the docker image `enqueue/dev` and push to docker hub

nick-zh:
seems there is more failing when trying to build the Dockerfile. I'll try to fix it, but i think i'll create a separate PR to not mix things up too much in here

nick-zh:
@makasim thx for merging the docker PR, give me a short shout when the image has been pushed to dockerhub, then i can merge master here and the tests should pass :v:

.travis.yml
```diff
@@ -68,6 +68,7 @@ install:
     - echo "memory_limit=2048M" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
     - php ./bin/fix-symfony-version.php "$SYMFONY_VERSION"
     - composer install
+    - sed -i 's/525568/16777471/' vendor/kwn/php-rdkafka-stubs/stubs/constants.php
```
nick-zh, line 71: this is needed for the tests, since the stub has a really old version

makasim: A comment near the line would be great.

nick-zh:
@makasim PR should be good to go now. The failing tests are G-PubSub related, from what i can see, this happened before my changes as well

PR #1102, opened 2020-10-05
[rdkafka] add non-blocking poll call to serve cb's

### LINKED ISSUE #749: [kafka] RdKafkaProducer has no way of error handling

nick-zh:
Should fix #749, calling `poll(0)` is non-blocking and poses minimal overhead. This way callbacks can be triggered, which is important to receive errors, delivery reports, etc. when producing a message.

### DIFF
--- a/pkg/rdkafka/RdKafkaProducer.php
+++ b/pkg/rdkafka/RdKafkaProducer.php
@@ -56,12 +56,14 @@ class RdKafkaProducer implements Producer
                 );
             } else {
                 $topic->producev($partition, 0 /* must be 0 */, $payload, $key, $message->getHeaders());
+                $this->producer->poll(0);
 
                 return;
             }
         }
 
         $topic->produce($partition, 0 /* must be 0 */, $payload, $key);
+        $this->producer->poll(0);
     }
 
     /**
--- a/pkg/rdkafka/Tests/RdKafkaProducerTest.php
+++ b/pkg/rdkafka/Tests/RdKafkaProducerTest.php
@@ -69,6 +69,11 @@ class RdKafkaProducerTest extends TestCase
             ->with('theQueueName')
             ->willReturn($kafkaTopic)
         ;
+        $kafkaProducer
+            ->expects($this->once())
+            ->method('poll')
+            ->with(0)
+        ;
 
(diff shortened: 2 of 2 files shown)


nick-zh:
@Steveb-p @makasim please have a look :v:

makasim (approved):

Steveb-p (approved):

Steveb-p:
Async nature of Kafka driver and since we're not really waiting for message production results errors might occur later on, but this is the least that can be done and should increase the overall stability ;)

@nick-zh thanks.

PR #1103, opened 2020-10-05
use supported librdkafka version of ext

nick-zh:
So `php-rdkafka:4.x` only supports `librdkafka:0.11.6` and up. I would strongly suggest though, that you force your users to use `librdkafka:1.x` and up. If you guys agree, i can create a PR to enforce this and adjust this one to use at least `librdkafka:1.0`
FYI: `php-rdkafka:5.0` will drop support for anything lower than `librdkafka:1.2`


nick-zh:
@Steveb-p @makasim please let me know what you think :v:

makasim:
Sounds good.

PR #1104, opened 2020-10-08
update image

nick-zh:
So i saw you guys were using `php:7.2`, in light that it is only supported until end of November, i took the liberty to fix it to 7.3, i hope that is ok. If not i can use the 7.2 tag :v: 
If you prefer to go straight to 7.4, let me know

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,18 +5,10 @@ RUN set -x && \
     apt-get update && \
     apt-get install -y --no-install-recommends --no-install-suggests wget curl openssl ca-certificates nano netcat php7.3-dev php7.3-redis php7.3-pgsql git python
 
-RUN set -x && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends --no-install-suggests php7.3-dev librabbitmq-dev make  && \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -1,32 +1,24 @@
...
     apt-get update && \
-    apt-get install -y --no-install-recommends --no-install-suggests wget curl openssl ca-certificates nano netcat php-dev php-redis php-pgsql git python
-
-RUN set -x && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends --no-install-suggests php-dev librabbitmq-dev make  && \
-    mkdir -p $HOME/php-amqp && \
-    cd $HOME/php-amqp && \
-    git clone https://github.com/pdezwart/php-amqp.git . && git checkout v1.9.3 && \
-    phpize --clean && phpize && ./configure && make install
```
nick-zh, line 14: from what i can see this is already installed in the formapro image and it is at `1.9.4`, if you want to move to the latest version `1.10.2` let me know, then i can add it in the `apt-get install` section

nick-zh: Done :smile:

makasim (approved):

Issue #1113, opened 2020-11-09
[ExceptionHandling] ProcessorExceptionExtensionInterface::onProcessorException not called for Errors


macghriogair:
# ProcessorExceptionExtensionInterface::onProcessorException is called only for Exceptions, not for Errors

## Expected

The QueueConsumer should catch all PHP `\Throwable` to allow for explicit error handling and controlled failure.
 
## Actual

Only \Exceptions are catched and passed to any extension implementing `ProcessorExceptionExtensionInterface`

## Context

I have an Extension for pushing poisoneous/failed messages to a dedicated failed jobs queue (the transport in use, beanstalkd, has no out of box support for so called dead letter queues) and realized that in case of PHP \Errors the Extension does not get called and the consume command exits without controlled message handling. This might be a serious problem not in suche way that the consumer exits (we use supervisord) but that the poisonous message gets stuck on top of the queue thus blocking all further message processing.

The problematic code is in `\Enqueue\Consumption\QueueConsumer` [line 198](https://github.com/php-enqueue/enqueue/blob/master/Consumption/QueueConsumer.php#L198) where it catches only `\Exception`, not `\Throwable`

```php

try {
    $result = $processor->process($message, $this->interopContext);
} catch (\Exception $e) {
    $result = $this->onProcessorException($extension, $consumer, $message, $e, $receivedAt);
}

```

PR #1114, opened 2020-11-09
[fix] queue consumption: catch throwable for processing errors

macghriogair:
This fixes the issue with unhandled Errors described in #1113

tbd. if this might be considered breaking backward compatibility or if I missed any important place.
Actually it should not, since \Exception extends from \Throwable. But feel free to comment & object.

### DIFF
--- a/pkg/enqueue/Consumption/QueueConsumer.php
+++ b/pkg/enqueue/Consumption/QueueConsumer.php
@@ -195,7 +195,7 @@ final class QueueConsumer implements QueueConsumerInterface
             if (null === $result) {
                 try {
                     $result = $processor->process($message, $this->interopContext);
-                } catch (\Exception $e) {
+                } catch (\Exception | \Throwable $e) {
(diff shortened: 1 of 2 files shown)


pkg/enqueue/Consumption/QueueConsumer.php
```diff
@@ -195,7 +195,7 @@ public function consume(ExtensionInterface $runtimeExtension = null): void
             if (null === $result) {
                 try {
                     $result = $processor->process($message, $this->interopContext);
-                } catch (\Exception $e) {
+                } catch (\Exception | \Throwable $e) {
```
makasim, line 198: not sure php 7.3 support it

macghriogair: I think it is there since PHP 7.1 https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.mulit-catch-exception-handling

makasim: Thanks for clarification

Issue #1108, opened 2020-10-25
PHP 8 Support?


markstory:
With PHP 8 coming out soon, I was wondering if you had started thinking about adding PHP8 to your CI matrix and composer constraint?

I'm happy to help with that process and can get it started, but I'm not familiar with all of the transports that enqueue supports and would likely require assistance to get everything working.

martinssipenko:
This is still relevant. @markstory is this something you are still willing to tackle? If yes then probably the first step should be to run CI on PHP 8.0 to see what happens and work from there.

markstory:
@martinssipenko I can start with a CI bump.

nicwortel:
Bad bot :robot:

danilopolani:
Bad bot again! But, for reference:

- https://github.com/php-enqueue/enqueue-dev/pull/1131
- https://github.com/php-enqueue/enqueue-dev/pull/1132

Steveb-p:
I spanked the bot and now he'll behave and not try to close this issue :police_officer:

PR #1116, opened 2020-11-24
Add PHP8 to CI builds

markstory:
Trying to get PHP8 compatibility started. This is likely wrong and broken, but need to start somewhere.

Refs #1108

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,10 @@ matrix:
       sudo: required
       services: docker
       env: SYMFONY_VERSION=4.3.* RDKAFKA_TESTS=true PREPARE_CONTAINER=true
+    - php: 8.0
+      sudo: required
+      services: docker
+      env: SYMFONY_VERSION=5.0.* FUNCTIONAL_TESTS=true PREPARE_CONTAINER=true
   allow_failures:
     - env: SYMFONY_VERSION=4.3.* RDKAFKA_TESTS=true PREPARE_CONTAINER=true
 
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
     "datadog/php-datadogstatsd": "^1.3"
   },
   "require-dev": {
-    "phpunit/phpunit": "^7.5",
+    "phpunit/phpunit": "^7.5|^8.5|^9.3",
     "phpstan/phpstan": "^0.12",
     "queue-interop/queue-spec": "^0.6",
     "symfony/browser-kit": "^3.4|^4",


svbackend:
Hello guys, could someone explain me the process of migration to php8? As far as I understand the main problem is phpunit update (from v7 to v9) as it's not backward compatible? So this enqueue package in order to support php8 should have a new branch with different php version requirements I suppose? Is there any branch where I can join to help migrate to php8? Because I just cloned this repository, but I can't just add a php^8 to composer.json and update travis.yml as all previous builds will fail because of new phpunit version (as you can see in this pull request).

Would be really nice to have some guidance/roadmap from maintainers with list of steps to accomplish in order to migrate to php8, I see it like this:
1. Create new branch for next major version of enqueue with php8 support
2. In this new branch travis.yml should be cleaned up, remove every php^7 build from there and add only new php8 builds 
3. Update tests to be compatible with phpunit 9
4. Update whole codebase, need to rename all newly reserved words (e.g class with name Match not longer allowed in php8, ref: https://github.com/sebastianbergmann/phpunit/issues/4533) 
5. ...
6. Profit!

is this sounds as a plan or am I missing something?

Steveb-p:
@svbackend in general yes, it boils down to updating phpunit dependency and checking if the codebase of library itself doesn't have any issues with PHP8 syntax.

One point I need to make though - this library will still need to maintain PHP7 compatibility for some time, at least until PHP 7.4 reaches it's end of life I suppose. This pretty much means that PHP8-specific syntax should not be used, as it will break compatibility with PHP7.

svbackend:
@Steveb-p ok, thanks, sounds clear, I will create a new PR within a week to get process started

Jurigag:
Anyone plans to do something here? It seems to me this PR died.

svbackend:
@Jurigag Yep, I'm working on it in different PR: 
https://github.com/php-enqueue/enqueue-dev/pull/1121

Hopefully to finish by the end of January, meanwhile I'm already using enqueue with redis on php 8, seems like it's working fine

Jurigag:
Then i think this PR could be closed then?

svbackend:
> Then i think this PR could be closed then?

I guess so

> Are you planning to also update related repos like https://github.com/php-enqueue/sns ?

It's monorepo, as far as I understand my changes here would be automatically applied to all other repositories for specific transports, so yes, sns also in that list

markstory:
I can close this given that #1121 is progressing better. I haven't had much time to dedicate to this sorry.

Issue #1117, opened 2020-12-02
[gps] Consuming message from external publisher in non-standard format


maciejzgadzaj:
I have a Symfony app with `messenger` using `enqueue/gps` to send/consume Pub/Sub messages. Everything works fine as long as messages are published and consumed by my app - they are in the format expected by `GpsMessage`, with `body`, `properties` and `headers` at the top level. 

But now I need to also consume messages sent by an external publisher, which do not adhere to this format - talking specifically about messages sent by Gmail API Push Notifications, which look like this: `{"emailAddress": "user@example.com", "historyId": "9876543210"}` (https://developers.google.com/gmail/api/guides/push#receiving_notifications) - no `body`, no `properties` and no `headers`. 

Trying to consume a message like this, `GpsMessage` gives me an error `Notice: Undefined index: body`, and fails.

Would appreciate any advice how can I make this work?

makasim:
I guess, the solution would be to make an attempt to parse message body and see weather it has `properties`, `headers`, `body` in it. 

If not, we can fallback and treat the body as body with empty properties and headers.

makasim:
here https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/gps/GpsConsumer.php#L111

maciejzgadzaj:
Ah, ok, thank you. Was hoping that it might be somehow doable with the current version. 

I'll look into necessary modifications.

maciejzgadzaj:
Ok, here's a [first draft](https://github.com/php-enqueue/enqueue-dev/pull/1118). I'm testing it now with my app, and setting up repo env to run tests etc, but so far it seems to work ok. Thoughts?

maciejzgadzaj:
Hi @makasim - is there anything preventing you from merging this?

makasim:
It looks good. Last thing is CI complaing about code style https://travis-ci.org/github/php-enqueue/enqueue-dev/jobs/748125221

maciejzgadzaj:
Yes, I've seen it, and I've chosen to ignore it - as it complains about the section which I haven't changed, and also, running the tests locally doesn't raise this issue at all. I guess we shouldn't really care about it here then?

makasim:
I understand that the CS issue is irrelevant to your change. Some times it happens. 

Could you please make CI happy ? I am happy to merge PR once CI green

maciejzgadzaj:
[...]

PR #1118, opened 2020-12-03
[gps] Add support for consuming message from external publisher in non-standard format

maciejzgadzaj:
See issue #1117


maciejzgadzaj:
Ok, I've added a couple of unit tests to for these 2 different scenarios, and now I'd say it's ready to merge if it's ok for you. That failed phpcs test I don't agree with (and local phpcs doesn't agree with it either) (plus, I haven't changed anything there either, so 🤷). Into your hands @makasim for a review!

PR #1131, opened 2021-01-25
PHPUnit 9.5

andrewmy:
- Update PHPUnit to 9.5;
- Drop PHP 7.1 and 7.2.

This is part of the effort to add support for PHP 8.


andrewmy:
@makasim take a look please :)

makasim (approved):

PR #1137, opened 2021-02-08
Implement against interfaces for rdkafka

TiMESPLiNTER:
As we wanted to extend some stuff from the rdkafka package we found out that most stuff is not really extendable. The classes are not marked as final but all properties are private so extending an Enqueue rdkafka class was quite useless.

Therefor I added proper interfaces for all rdkafka related classes so one can decorate them properly. All methods should now use interfaces as parameters (either Interop or Enqueue/Rdkafka ones).

Not sure what's the reason why there are no interfaces and not sure about the naming of the interfaces itself.

Let me know what you think.


makasim:
What kind of stuff do you want to extend ?

TiMESPLiNTER:
@makasim We needed to make it possible to subscribe to a specific offset/partition with one of our consumers.

makasim:
Is it wide-needed feature ? Maybe we can implement it here?

TiMESPLiNTER:
I don't know if it's a wide needed feature honestly. It's just a possibility that Kafka offers. We could add it to the Enqueue rdkafka package as well. But first it would be great to prepare the package to provide better extensibility capabilities that's what this PR is for. 

I would suggest to create another PR for adding support to subscribe to a specific offset if it's something we want to include into this package as well.

makasim:
Is it what you want https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/RdKafkaConsumer.php#L69 ?

TiMESPLiNTER:
@makasim exactly. But we needed it to be set here via `$config` (like `commit_async`) -> https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/RdKafkaContext.php#L138

makasim:
That would force exact same partition and offset for all consumer and different topics. Is that what you want ?

TiMESPLiNTER:
As we only have one partition and one consumer: yes, that's what we wanted in this particular case. As I mentioned earlier it were "difficult circumstances". It was the best solution we were able to come up with because we use Enqueue together with the symfony/messenger component and sroze's bridge.

PR #1138, opened 2021-02-09
Fix package CI

andrewmy:
- Add push to master trigger;
- Replace the clunky cache + composer install steps with an Action;
- Replace Travis with GH WF in packages;
- Add the missing CI to amqp-tools;
- Replace the shields.


andrewmy:
@makasim not sure how to test every package's CI here but take a look.
Especially concerned about gearman, maybe some another package would require an additional PR to fix it.

makasim:
> Especially concerned about gearman, maybe some another package would require an additional PR to fix it.

That's ok. We'll see

andrewmy:
Ok, thanks for the merge, now you got to enable GH workflows on all the split repos :)

PR #1139, opened 2021-02-11
Split tests into different matrices; fix highest/lowest dependencies

andrewmy:
The original build matrix looked quite funky, trying to make it more understandable and expandable here.﻿

Also recently I encountered bugs in a different library which had quite wide lowest/highest deps gap, made me think we should test it here as well.

- Split tests into different matrices;
- Add highest/lowest dependencies;
- Fix lowest dependency versions;
- Drop Symfony 4.3 and 5.0 tests;
- Fix composer install race condition in CI.


andrewmy:
Turns out bunny 0.2 had open php req and "allowed" php 8 🙈 Discovered it while bumping react libs to ^1.0.
Had to fix the bunny requirement here which duplicates #1143 — happy to have it merged first if it happens so.

andrewmy:
The matrix is huge :o
Symfony 4.3 and 5.0 are EOL — do we want to test against them? See https://symfony.com/releases

andrewmy:
@makasim @Steveb-p now the PR is ready to go, sorry for not making it a draft initially.
I've still got a question about Symfony 4.3 though, see above.

makasim:
> Symfony 4.3 and 5.0 are EOL — do we want to test against them? See https://symfony.com/releases

I dont think so.

andrewmy:
Looks ready to me now, and also fixes the composer install race condition which may impact other PRs

makasim (approved):

PR #1143, opened 2021-02-16
Allow bunny 0.5

snapshotpl:


pkg/amqp-bunny/composer.json
```diff
@@ -9,7 +9,7 @@
         "php": "^7.3",
         "queue-interop/amqp-interop": "^0.8",
         "queue-interop/queue-interop": "^0.8",
-        "bunny/bunny": "^0.4",
+        "bunny/bunny": "^0.4 || ^0.5",
```
andrewmy, line 12: I suggest adding the same to the root composer.json to run all the checks

makasim:
@snapshotpl there are some conflicts. Could you take a look ?

makasim:
good) sorry for bothering you

Issue #1022, opened 2020-02-12
CreateTopic is called


hpatoio:
I've just installed and configured enqueue. Why is `CreateTopic` called ? I already created the SNS topic but enqueue call it anyway.

Steveb-p:
`CreateTopic` does not actually create topic on broker, but instead it creates an object responsible for working with a topic / queue.

At least that's my assumption, since I'm not familiar with SNS/SQS - but almost every enqueue package behaves this way.

Unless it enqueue actually tries to create a topic on a broker and causes you an issue, I'm closing this.

tyloo:
@Steveb-p : It actually does a `createTopic` query to the broker. Problem is, we have set our IAM credentials to not have the right to create a topic.

> AuthorizationError (client): User: arn:aws:sts::... is not authorized to perform: SNS:CreateTopic on resource: arn:aws:sns:eu-west-1:...:dev-messenger

tyloo:
@Steveb-p : It actually does a `createTopic` query to the broker. Problem is, we have set our IAM credentials to not have the right to create a topic.

> AuthorizationError (client): User: arn:aws:sts::... is not authorized to perform: SNS:CreateTopic on resource: arn:aws:sns:eu-west-1:...:dev-messenger

jderusse:
Actually, SNS component creates a topic in order to fetch the ARN:

https://github.com/php-enqueue/sns/blob/0568fa5168b39a48362a46db5638a5d0f586632f/SnsContext.php#L66-L77

which is called in 
https://github.com/php-enqueue/sns/blob/0568fa5168b39a48362a46db5638a5d0f586632f/SnsContext.php#L144-L151

Please reopen the issue.

timgregg:
@Steveb-p It does create an sns topic is it possible to reopen this issue

makasim:
fixed in #1147

PR #1147, opened 2021-03-18
[sns] added possibility to define already existing topics (prevent create topic call) #1022

paramonov:


pkg/sns/SnsContext.php
```diff
@@ -31,12 +31,11 @@ class SnsContext implements Context
...
     private $topicArns;
 
-    public function __construct(SnsClient $client, array $config)
+    public function __construct(SnsClient $client, array $config, array $topicArns = [])
     {
         $this->client = $client;
         $this->config = $config;
-
-        $this->topicArns = [];
+        $this->topicArns = $topicArns;
```
makasim, line 38: topicArns property format is 

```
topic_name => arn 
```

pkg/sns/SnsConnectionFactory.php
```diff
@@ -71,7 +72,7 @@ public function __construct($config = 'sns:')
      */
     public function createContext(): Context
     {
-        return new SnsContext($this->establishConnection(), $this->config);
+        return new SnsContext($this->establishConnection(), $this->config, $this->config['topic_arns']);
```
makasim, line 75: why do you pass topc_arns as separate arg? They are already there in config which is passed as second arg

paramonov: removed third arg

pkg/sns/SnsContext.php
paramonov: yep, refactored

pkg/sns/SnsConnectionFactory.php
```diff
@@ -59,6 +60,9 @@ public function __construct($config = 'sns:')
 
                 unset($config['dsn']);
             }
+            if (\array_key_exists('topic_arns', $config) && \is_string($config['topic_arns'])) {
```
makasim, line 63: at this stage it has to be assoc array already. The string parsing should happen if dsn is passed.

pkg/sns/Tests/SnsConnectionFactoryConfigTest.php
```diff
@@ -148,6 +154,24 @@ public static function provideConfigs()
                 'version' => '2010-03-31',
                 'lazy' => false,
                 'endpoint' => 'http://localstack:1111',
+                'topic_arns' => [],
+            ],
+        ];
+
+        yield [
+            ['dsn' => 'sns:?topic_arns=topic1|arn:aws:sns:us-east-1:123456789012:topic1;topic2|arn:aws:sns:us-west-2:123456789012:topic2'],
```
makasim, line 162: what about passing arns as

```
topic_arns[topic1]=arn:aws:sns:us-east-1:123456789012&topic_arns[topic2]=arn:aws:sns:us-west-2:123456789012:topic2
```

that way the standard query parsing function does the job

paramonov: good point, replaced

pkg/sns/SnsConnectionFactory.php
paramonov: removed, not needed when using dsn parsing

pkg/sns/SnsContext.php
```diff
@@ -35,8 +35,7 @@ public function __construct(SnsClient $client, array $config)
     {
         $this->client = $client;
         $this->config = $config;
-
-        $this->topicArns = [];
+        $this->topicArns = $config['topic_arns'];
```
makasim, line 38: this is BC break. set empty array if not set

makasim:
fixes #1022

PR #1150, opened 2021-03-24
Allow to configure a Serializer for Kafka

coreation:
## Notes

Hello everyone, first off, great job on making this such an active PHP project!
I'd like to use the Kafka queue, however my messages are encoded in Avro, which means I need to serialise them using a Schema registry. The only clean way without being intrusive, according to me, was to add a configuration entry called "serializer" and to capture this when setting up the RdKafkaContext.

If there are better ways of doing this, I'd love to hear them, if I need to alter some code or extra checks, i.e. check if the passes class exists before trying to instantiate it, let me know!

**UPDATE**
It was not my intention to edit the composer.json file, but it seems that the master version where I forked from has the requirement of needing "all" of the php extensions for all of the different queues this package supports.

I think I need some pointers on how to contribute best to this package, as the rdkafka is a read-only one.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,6 @@
     "google/cloud-pubsub": "^1.4.3",
     "doctrine/orm": "^2.7.4",
     "doctrine/persistence": "^1.3.3|^2.0",
-    "mongodb/mongodb": "^1.2",
     "aws/aws-sdk-php": "^3.155",
     "stomp-php/stomp-php": "^4.5|^5",
     "php-http/guzzle7-adapter": "^0.1.1",
--- a/composer.json
+++ b/composer.json
@@ -17,11 +17,6 @@
   "require": {
     "php": "^7.3|^8.0",
 
-    "ext-amqp": "^1.9.3",
-    "ext-gearman": "^2.0",
-    "ext-mongodb": "^1.5",
-    "ext-rdkafka": "^4.0|^5.0",
-
     "queue-interop/amqp-interop": "^0.8.2",
     "queue-interop/queue-interop": "^0.8.1",
     "bunny/bunny": "^0.4|^0.5",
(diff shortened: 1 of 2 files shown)


makasim:
It should be already possible to do. 

You must implement Serializer interface (https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/Serializer.php) and pass it to the context (https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/RdKafkaContext.php#L58).

makasim:
I dont think that passing serializer in a config is a good idea. It should be set at runtime.

coreation:
Thanks for the reply @makasim I'm using this in combination with the Enqueue Laravel package, how do you see this working "at runtime" exactly? The only way I see this being possible is to pass a serializer class path in the queue config. 

Am I correct to say that if I'd need this package, I would need to install all of the extensions (amqp, mongodb, ...) even though I just need one of the driver queues in this package?

coreation:
> It should be already possible to do.
> 
> You must implement Serializer interface (https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/Serializer.php) and pass it to the context (https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/rdkafka/RdKafkaContext.php#L58).

I think in my PR that's what I'm doing, except for the fact that right now, it's being hard coded as a JsonSerializer, it first checks if it needs to use a different serializer. 

I just wanted to add that this PR isn't ready to be merged as I had to remove several dependencies of queue drivers that I don't need in my own branch. So if you could tell me how to handle that better (cfr. my previous comment), that'd be great! This package helped me save some work so anything I can do to contribute I'd be happy to.

Issue #1151, opened 2021-03-24
enqueue/job-queue comppatibility with doctrine/orm 2.8


GothShoot:
Hello,

I am interested in using the package enqueue/job-queue in a project, but it is powered by PHP 8 and needs doctrine 2.8 to run on PHP 8.

It is possible to make it compatible with this version of ORM?

Thanks.

makasim:
Lookf for some older versions (prior PHP8).

GothShoot:
I can try for support doctrine 2.8 in a PR ?
Client already in PHP 8 in production.

Steveb-p:
@GothShoot definitely. There should be little or no changes related to PHP8 itself, and especially if you already have a patched version for yourself.

GothShoot:
I try to support 👍

GothShoot:
See #1155 👷‍♂️

PR #1155, opened 2021-03-25
Try to change doctrine/orm version for supporting 2.8 (PHP 8 support).

GothShoot:
Try to use doctrine/orm 2.8 with job queue, this is the last step to php 8 support.

### DIFF
--- a/pkg/job-queue/composer.json
+++ b/pkg/job-queue/composer.json
@@ -10,7 +10,7 @@
         "enqueue/enqueue": "^0.10",
         "enqueue/null": "^0.10",
         "queue-interop/queue-interop": "^0.8",
-        "doctrine/orm": "~2.7.4",
+        "doctrine/orm": "~2.7.4 | ^2.8",
         "doctrine/dbal": "^2.12"
     },
     "require-dev": {


makasim:
Have you tested it ?

GothShoot:
Need to setup a work environment because it needs env var to work with your testing config.
I give you a response when I have tested 👍

makasim:
You should not require this repo directly. fork read only package, apply your changes and require it in composer. 

Once done, backport changes to this repo.

GothShoot:
My fork run on my app (just installing and clear the cache)

makasim:
Good, thanks

pkg/job-queue/composer.json
```diff
@@ -10,7 +10,7 @@
         "enqueue/enqueue": "^0.10",
         "enqueue/null": "^0.10",
         "queue-interop/queue-interop": "^0.8",
-        "doctrine/orm": "~2.7.4",
+        "doctrine/orm": "~2.7.4 | ^2.8",
```
deguif, line 13: Can not this be simplified to `"^2.7.4"` only?

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/1151

PR #1157, opened 2021-04-02
register worker callback only once, move to constructor

cturbelin:
New PR with clean history following discussion in #1141


makasim:
Thanks!

makasim:
fixes https://github.com/php-enqueue/enqueue-dev/issues/1140

PR #1159, opened 2021-04-21
add subscription_interval as config for dbal subscription consumer

mordilion:
I am facing an issue with the situation, that I have a few JobWorker running and my DB doesn't like that many requests. Without the change, I am getting 50 Requests in 10 Seconds with 1 JobWorker. Because of this, I am getting many "MySQL Server has gone away" messages.

### DIFF
--- a/pkg/dbal/DbalSubscriptionConsumer.php
+++ b/pkg/dbal/DbalSubscriptionConsumer.php
@@ -37,6 +37,13 @@ class DbalSubscriptionConsumer implements SubscriptionConsumer
     private $redeliveryDelay;
 
     /**
+     * Time to wait between subscription requests in milliseconds.
+     *
+     * @var int
+     */
+    private $subscriptionInterval = 200;
+
+    /**
      * @param DbalContext $context
      */
     public function __construct(DbalContext $context)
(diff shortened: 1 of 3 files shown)


mordilion:
@makasim is it possible to get at least a short feedback for my Pull-Request? Right now I am patching my Code every time to get that option. At least a "No" would be nice, because then I could start looking for another Library.

In my opinion, it would make sense to add that option. :)

pkg/dbal/DbalSubscriptionConsumer.php
```diff
@@ -36,6 +36,13 @@ class DbalSubscriptionConsumer implements SubscriptionConsumer
      */
     private $redeliveryDelay;
 
+    /**
+     * Time to wait between subscription requests in milliseconds.
+     *
+     * @var int
+     */
+    private $subscriptionInterval = 200;
```
makasim, line 44: What about naming it pollingInterval like [here](https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/fs/FsConsumer.php#L39) to stay consistent with other transports ? 

The tests should be fixed

makasim (changes requested):

makasim:
In general it looks good.

mordilion (commented):
Changed the naming :)

Issue #1163, opened 2021-04-23
Support Connection name


Webonaute:
https://github.com/supersmile2009/amqp-tools/commit/0742ee44a5bb118dd656ffc070f686da081f0ffa

is it possible to merge this patch so we can get get rid of this patch in our composer?
We use this since a year now to know which consumer is connected with which version of our api.

makasim:
The patch looks good to me. I'll merge it once someone open PR on enqueue-dev repo.

Webonaute:
here it is

PR #1164, opened 2021-04-23
Support connection_name

Webonaute:
Related to #1163


makasim:
Tests have to be fixed.

PR #1161, opened 2021-04-23
Fix dbal delivery delay to always keep integer value

dgafka:
Fixes incorrect calculations that may result in float instead of int:

`SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for integer: \"1619173911.5\" at /data/app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractPostgreSQLDriver.php:102, Doctrine\\DBAL\\Driver\\PDO\\Exception(code: 22P02): SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for integer: \"1619173911.5\" at /data/app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDO/Exception.php:18, PDOException(code: 22P02): SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for integer: \"1619173911.5\" at /data/app/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:115)`

PR #1167, opened 2021-04-29
Fix dbal delivery delay to always keep integer value

dgafka:
In this issue I have fixed TTL: https://github.com/php-enqueue/enqueue-dev/pull/1161
This MR fixes the delay.


Steveb-p (approved):

dgafka:
@msheakoski can tag for the enqueue/dbal be released with this change?

Issue #1166, opened 2021-04-29
Support DBAL v3.0


m3m0r7:
We use the DBAL v3, but I'm thinking https://github.com/php-enqueue/dbal is not supported DBAL v3.
Could you support or have you plan to upgrade DBAL v3?

makasim:
PR is welcome.

m3m0r7:
OK, I will make a PR

PR #1169, opened 2021-05-02
Support dbal v3

m3m0r7:
Support dbal v3 #1166


pkg/dbal/DbalTypeResolverTrait.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Enqueue\Dbal;
+
+trait DbalTypeResolverTrait
+{
+    protected static function resolveDbalType(string $typeName): string
```
m3m0r7, line 9: The method is compatible for dbal v2 and v3

Steveb-p: This makes it a requirement to include in all classes that need to use the compatibility layer between DBAL v2 & v3.

Considering that it would be a static method, there is really no need for it. It can be made a static method of a class on it's own - it does not depend on the class you want to attach it at all.

It also prevents any type of static code analysis from picking up potential errors (previously you had a well defined constant list, now you can't use that because you're depending on reflection to find class constants.

What I'd suggest instead is a "class with side effects". Normally you'd avoid creating classes that change their structure depending on the other libraries, but this would be a compatibility layer that is supposed to be adapting to DBAL version.

It would go like this:

```php
if (class_exists(\Doctrine\DBAL\Types\Types::class)) {
    class DbalType {
        // constants defined here
    }
} else {
    class DbalType {
        // constants from the other version
    }
}
```

Also note that you don't need to use class name as a string. You can use the `::class` notation, because **using class name / type on it's own does not trigger autoloading**.

m3m0r7: @Steveb-p 
Thanks for your reviews. I will fix it.

> using class name / type on it's own does not trigger autoloading.

I did not know its feature, Thanks. I thought it will be shown an error.

Steveb-p: > I did not know its feature, Thanks. I thought it will be shown an error.

Yup. PHP is smart enough to figure out what class you're talking (taking into account current namespace and imports) and doesn't bother with autoloading until it's actually necessary (usually when creating a new instance, calling a static method or passing it to `class_exists` function and it's "family").

This has the side-effect of also not triggering autoload when doing `$x instanceof FooBar` (`FooBar` is not autoloaded), and the same applies to type-hints (even method arguments / property types!). In 99.9% of cases it's not relevant, as at some point you had to have an instance somewhere before (so the class is loaded). This becomes an issue (that 0.01% :smile: ) when you're using `class_alias`, but that's really a special case.

See https://www.schmengler-se.de/en/2016/09/php-using-class_alias-to-maintain-bc-while-move-rename-classes/

m3m0r7: Sorry for the late commit 🙏  I did fix it.
I'm thinking that do not necessary to use the if statement when creating a new class and copy&paste from the dbal source.
Therefore,

PR #1171, opened 2021-05-18
Fixes timeout in messages

Csardelacal:
My understanding is that the old code would have an exponential decay. Basically, it would subtract the amount of seconds that had elapsed since the code started executing with each attempt.

This would mean that a retry after two seconds would subtract 2 seconds from the timeout, a retry after 5 seconds, would remove another 5 seconds.

### DIFF
--- a/pkg/redis/RedisConsumerHelperTrait.php
+++ b/pkg/redis/RedisConsumerHelperTrait.php
@@ -34,7 +34,7 @@ trait RedisConsumerHelperTrait
         while ($expires > time()) {
             $this->migrateExpiredMessages($this->queueNames);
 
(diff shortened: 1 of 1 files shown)


Csardelacal:
Is there an issue with this code?

blondie101010 (commented):
While I think your solution is much better and does make the process more sane, this change could cause some side effects from code expecting the current behavior.

Csardelacal:
I don't see an issue with existing implementations coming into play. If somebody requested a timeout of 60 seconds for their application the only side-effect of this code would be that the application is a bit more patient. Since the old timeout would decay too fast. It may slow somebody's application down, but it's definitely not going to cause it to behave poorly.

Issue #1156, opened 2021-03-30
EnqueueBundle: ResetServicesExtension needs a new message to perform reset


djmarland:
https://github.com/php-enqueue/enqueue-dev/blob/master/pkg/enqueue-bundle/Consumption/Extension/ResetServicesExtension.php

This extension implements `MessageReceivedExtensionInterface` so it runs when a new message is received.
When using the FingersCrossed Monolog handler this means that it requires a second message before logs will be output. If there is a time gap before the next message some important logs could be missed.

Would this extension be better off running _after_ a message is processed, such as implementing `PostMessageReceivedExtensionInterface` instead?

makasim:
yeap, that makes sense to me.

djmarland:
PR opened and linked

PR #1172, opened 2021-05-24
ResetServicesExtension to run _after_ message received

djmarland:
Resolves https://github.com/php-enqueue/enqueue-dev/issues/1156

Reset the services after message is received, so that logs are flushed without having to wait for a new message


djmarland:
Thank you

makasim:
welcome

PR #1174, opened 2021-05-25
Changed: Clear Doctrine cache before calculate Root Job Status

aigarszuika:
Before determine Root Job Status Doctrine cache for Job entity must be cleared, otherwise if this root job was processed in previous request by the same enqueue, then it will falsely return old status, which will cause infinite running root jobs


aigarszuika:
It is safer to refresh each child entity when performing root job status check.

ksd-krauzand (approved):

PR #1182, opened 2021-07-04
Remove call deprecated method "setLogLevel"

guil95:
Remove the deprecated "setLogLevel" method call, to avoid throwing an exception


andrewmy:
Hey @guil95 the master branch now has fixed function test pipeline, please rebase or update your branch to take advantage of this ✌️

PR #1196, opened 2021-08-24
Fix partition's choice for the cases when partition number is zero

rodrigosarmentopicpay:


makasim:
could you add a test forthis case ?

rodrigosarmentopicpay:
> could you add a test forthis case ?

Ok, I will do it.

rodrigosarmentopicpay:
> could you add a test forthis case ?

The tests are included as requested.

rodrigosarmentopicpay:
Hi Max, is there anything missing? The error seems to be related to the rabbitmq repo, not to my changes.

makasim:
It is, yes.  But unfortunately  I can merge only green PR. So the issue has to be addressed.

rodrigosarmentopicpay:
> It is, yes. But unfortunately I can merge only green PR. So the issue has to be addressed.

Ok, now it's all ok!

makasim:
@rodrigosarmentopicpay thank you, merged

PR #1201, opened 2021-09-30
No waiting between tube switch for dbal subscription

mordilion:
I am facing the issue, that when I have multiple queues in the System but just one has jobs, the usleep happens between each job.

Example:
Queues: 
    - JobQueue1 -> 0 jobs
    - JobQueue2 -> 2000 jobs
    - JobQueue3 -> 0 jobs

If the subscription_polling_interval is set to 2 seconds, then it will wait 2 seconds until it will start with the next job from the same queue.

I have added the logic to use the order of the array to select the next job ordered by the array order. After a job is done I rearrange the array to put the queue from the last job at the end of the array.


mordilion:
This is actually less performant. Please have a look at https://github.com/php-enqueue/enqueue-dev/pull/1202

PR #1202, opened 2021-09-30
Disable Subscription-Polling Sleep while queue items available 

mordilion:
This is a more performant solution for the PR https://github.com/php-enqueue/enqueue-dev/pull/1201

It just works until there is no more message in the queue. The priority and published_at are the key for what will be next worked on.


jprangenbergde:
That would be great!

jprangenbergde:
First in, First out is in my opinion the correct way. Same priority? > published_at. Why should a job „A“ which has the same priority like job „B“ but got published After job „B“ get executed before? First in, first out on same priority is the way to go.

mordilion:
@makasim would be the current solution fine for you? I changed the logic in that way, that it goes through the queues now.

mordilion:
@makasim I would like to come back to this topic. The issue which I have is that the SubscriptionPollingDelay is happening between the change of the queue and not only if all queues are empty. My current solution should still fit your requirements to loop through the queues but not cause the issue which I am currently facing.

mordilion:
@makasim I changed the main code back to the original one and added a check if a queue was consumed. I a queue was consumed, then it works further without waiting because we expect more jobs. If no queue was processed, then it waits for the configured polling interval.

PR #1208, opened 2021-10-25
Set to Header only Headers

s1lver:
I`m setting headers and properties

Снимок экрана 2021-10-25 в 17 13 04

If i use function ```$message->getHeaders()``` i get properties instead and if use ```$message->getProperties()``` i get headers!!!

This PR fixing it

**Before fix**
```json
{"headers":{"delivery_mode":1,"user_id":"guest"}}
```

**After fix**
```json
{"headers":{"traceparent":"00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01","z-traceparent":"dsjdhsjdhsjds"}}
```

### DIFF
--- a/pkg/amqp-lib/AmqpContext.php
+++ b/pkg/amqp-lib/AmqpContext.php
@@ -299,14 +299,14 @@ class AmqpContext implements InteropAmqpContext, DelayStrategyAware
      */
     public function convertMessage(LibAMQPMessage $amqpMessage): InteropAmqpMessage
     {
-        $headers = new AMQPTable($amqpMessage->get_properties());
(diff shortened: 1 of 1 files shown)


makasim:
This would be a huge BC break. There are a lot of projects which rely on current behavior, even though it's not correct.

s1lver:
> This would be a huge BC break. There are a lot of projects which rely on current behavior, even though it's not correct.

Maybe add a description to UPGRADE.md?

makasim:
That would require a version bump and a lot of work on my side. I tend to avoid it.

PR #1225, opened 2021-12-13
Fix types for Symfony 5.4

shyim:
image


ruudk:
@shyim Thanks for this. Could you please also fix the code style issue?

shyim:
@ruudk done

ruudk:
@makasim Could you approve the workflow so that we can see if it passes?

shyim:
@ruudk I guess the failing one fails also in master

ruudk:
@makasim What do you think, can this be merged?

shyim:
Can we have this released, so we can get rid of that error?

PR #1228, opened 2022-01-06
Compatiable with psr/container

FuriosoJack:
With compatible with interface container psr


bin/patch-type-declarations
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    if (function_exists('stream_wrapper_register') && stream_wrapper_register('phpvfscomposer', 'Composer\BinProxyWrapper')) {
+        include("phpvfscomposer://" . __DIR__ . '/..'.'/vendor/symfony/error-handler/Resources/bin/patch-type-declarations');
+        exit(0);
+    }
+}
+
+include __DIR__ . '/..'.'/vendor/symfony/error-handler/Resources/bin/patch-type-declarations';
```
ruudk, line 107: What's the purpose of this file? Looks like you committed this by accident.

FuriosoJack: Woow yes, I will create another commit to remove it-

FuriosoJack: > What's the purpose of this file? Looks like you committed this by accident.

but you know I think that this file is generated when I execute any of the commands that you have in the contribution document

ruudk: Then the file should probably be added to the .gitignore file.

Please rebase so that there is only 1 commit.

PR #1230, opened 2022-01-11
#1229 Allow return null in kafka Serializer (delete messages aka Tombstones)

### LINKED ISSUE #1229: Impossible send null payload (tombstone) because interface Serializer do not allow return null
Payload to send [defined by serializer](https://github.com/php-enqueue/enqueue-dev/blob/66f15d0704a21271dc51d0ef0cbcdfb61a8ed940/pkg/rdkafka/RdKafkaProducer.php#L41) in the [RdKafkaProducer::send](https://github.com/php-enqueue/enqueue-dev/blob/66f15d0704a21271dc51d0ef0cbcdfb61a8ed940/pkg/rdkafka/RdKafkaProducer.php#L35):
```php
        $payload = $this->serializer->toString($message);
```
Where serializer is the [instance of a class, implementing](https://github.com/php-enqueue/enqueue-dev/blob/66f15d0704a21271dc51d0ef0cbcdfb61a8ed940/pkg/rdkafka/RdKafkaProducer.php#L28) [Enqueue\RdKafka\Serializer](https://github.com/php-enqueue/enqueue-dev/blob/66f15d0704a21271dc51d0ef0cbcdfb61a8ed940/pkg/rdkafka/Serializer.php#L7) interface. There method [toString](https://github.com/php-enqueue/enqueue-dev/blob/66f15d0704a21271dc51d0ef0cbcdfb61a8ed940/pkg/rdkafka/Serializer.php#L9) has signature:
```php
    public function toString(RdKafkaMessage $message): string;
```

So, in my class, I just can't return null! Please change the return type to `?string`.

Using `nulls` as `payload` is [common practice also known as tombstones](https://medium.com/@damienthomlutz/deleting-records-in-kafka-aka-tombstones-651114655a16).

Hubbitus:
Closes #1229

### DIFF
--- a/pkg/rdkafka/JsonSerializer.php
+++ b/pkg/rdkafka/JsonSerializer.php
@@ -6,7 +6,7 @@ namespace Enqueue\RdKafka;
 
 class JsonSerializer implements Serializer
 {
-    public function toString(RdKafkaMessage $message): string
+    public function toString(RdKafkaMessage $message): ?string
     {
         $json = json_encode([
             'body' => $message->getBody(),
@@ -14,26 +14,22 @@ class JsonSerializer implements Serializer
             'headers' => $message->getHeaders(),
         ]);
 
-        if (JSON_ERROR_NONE !== json_last_error()) {
-            throw new \InvalidArgumentException(sprintf(
-                'The malformed json given. Error %s and message %s',
-                json_last_error(),
-                json_last_error_msg()
-            ));
+        if (\JSON_ERROR_NONE !== json_last_error()) {
+            throw new \InvalidArgumentException(sprintf('The malformed json given. Error %s and message %s', json_last_error(), json_last_error_msg()));
         }
 
         return $json;
     }
 
-    public function toMessage(string $string): RdKafkaMessage
+    public function toMessage(?string $string): RdKafkaMessage
     {
-        $data = json_decode($string, true);
-        if (JSON_ERROR_NONE !== json_last_error()) {
-            throw new \InvalidArgumentException(sprintf(
-                'The malformed json given. Error %s and message %s',
-                json_last_error(),
-                json_last_error_msg()
-            ));
+        if (null !== $string) {
+            $data = json_decode($string, true);
+            if (\JSON_ERROR_NONE !== json_last_error()) {
+                throw new \InvalidArgumentException(sprintf('The malformed json given. Error %s and message %s', json_last_error(), json_last_error_msg()));
+            }
+        } else {
+            $data = ['body' => null, 'properties' => null, 'headers' => 'headers'];
         }
 
         return new RdKafkaMessage($data['body'], $data['properties'], $data['headers']);
--- a/pkg/rdkafka/JsonSerializer.php
+++ b/pkg/rdkafka/JsonSerializer.php
@@ -23,13 +23,13 @@ class JsonSerializer implements Serializer
 
     public function toMessage(?string $string): RdKafkaMessage
     {
-        if (null !== $string) {
-            $data = json_decode($string, true);
-            if (\JSON_ERROR_NONE !== json_last_error()) {
-                throw new \InvalidArgumentException(sprintf('The malformed json given. Error %s and message %s', json_last_error(), json_last_error_msg()));
-            }
-        } else {
-            $data = ['body' => null, 'properties' => null, 'headers' => 'headers'];
+        if (null === $string) {
+            return new RdKafkaMessage(null, null, null);
+        }
+
+        $data = json_decode($string, true);
+        if (\JSON_ERROR_NONE !== json_last_error()) {
+            throw new \InvalidArgumentException(sprintf('The malformed json given. Error %s and message %s', json_last_error(), json_last_error_msg()));
         }
 
         return new RdKafkaMessage($data['body'], $data['properties'], $data['headers']);
--- a/pkg/rdkafka/Serializer.php
+++ b/pkg/rdkafka/Serializer.php
@@ -6,7 +6,7 @@ namespace Enqueue\RdKafka;
 
 interface Serializer
 {
-    public function toString(RdKafkaMessage $message): string;
+    public function toString(RdKafkaMessage $message): ?string;
 
-    public function toMessage(string $string): RdKafkaMessage;
+    public function toMessage(?string $string): RdKafkaMessage;
 }


Hubbitus:
There are some errors in automated tests like:
```

1) Enqueue\Wamp\Tests\Functional\WampConsumerTest::testShouldSendAndReceiveMessage
Error: Call to undefined function GuzzleHttp\Psr7\uri_for()

/mqdev/vendor/ratchet/pawl/src/Connector.php:126
/mqdev/vendor/ratchet/pawl/src/Connector.php:38
/mqdev/vendor/thruway/pawl-transport/src/PawlTransportProvider.php:59
/mqdev/vendor/thruway/client/src/Peer/Client.php:230
/mqdev/pkg/wamp/WampConsumer.php:96
/mqdev/pkg/wamp/Tests/Functional/WampConsumerTest.php:36
/mqdev/pkg/test/RetryTrait.php:26
phpvfscomposer:///mqdev/vendor/phpunit/phpunit/phpunit:97
```
But that looks like can't be introduced by my change. Could it be possible just restart failed actions?

Steveb-p:
Seems like too high version of `guzzle/psr7` is used (2.1.0) and we are using some deprecated functions - or rather, they were not upgraded.

See https://github.com/guzzle/psr7#upgrading-from-function-api.

And yeah, it's outside of the scope of change.

Hubbitus:
@Steveb-p, thank you for the answer.

> And yeah, it's outside of the scope of change.

So, could you please review the change itself? That is very simple I think.

pkg/rdkafka/JsonSerializer.php
```diff
@@ -6,34 +6,30 @@
...
-                json_last_error(),
-                json_last_error_msg()
-            ));
+        if (null !== $string) {
+            $data = json_decode($string, true);
+            if (\JSON_ERROR_NONE !== json_last_error()) {
+                throw new \InvalidArgumentException(sprintf('The malformed json given. Error %s and message %s', json_last_error(), json_last_error_msg()));
+            }
+        } else {
+            $data = ['body' => null, 'properties' => null, 'headers' => 'headers'];
```
Steveb-p, line 32: Prefer early return for readability:
```suggestion
        if (null === $string) {
            return new RdKafkaMessage(null, null, null);
        }
        
        $data = json_decode($string, true);
        if (\JSON_ERROR_NONE !== json_last_error()) {
            throw new \InvalidArgumentException(sprintf('The malformed json given. Error %s and message %s', json_last_error(), json_last_error_msg()));
        }
```

Hubbitus: Ok, applied.

pkg/rdkafka/JsonSerializer.php
```diff
@@ -6,34 +6,30 @@
 
 class JsonSerializer implements Serializer
 {
-    public function toString(RdKafkaMessage $message): string
+    public function toString(RdKafkaMessage $message): ?string
```
Steveb-p, line 9: That's not true, and it does not match with the `Stringable` interface that is part of PHP8.

See https://www.php.net/manual/en/class.stringable.php.

```suggestion
    public function toString(RdKafkaMessage $message): string
```

pkg/rdkafka/Serializer.php
```diff
@@ -6,7 +6,7 @@
 
 interface Serializer
 {
-    public function toString(RdKafkaMessage $message): string;
+    public function toString(RdKafkaMessage $message): ?string;
```
Steveb-p, line 9:
```suggestion
    public function toString(RdKafkaMessage $message): string;
```

Steveb-p (commented):
`toString` should always return a string, especially since it's based on `RdKafkaMessage` and can be thought of as an implementation of `Stringable`.

Otherwise I have no further comments, and it looks good to me.

pkg/rdkafka/JsonSerializer.php
Hubbitus: But `Serializer` does not extend `Stringable`?

Hubbitus: And [Stringable](https://www.php.net/manual/en/class.stringable.php) defines interface `public __toString(): string` method with two underscores in name, not `toString`

Hubbitus:
@Steveb-p, [`Stringable`](https://www.php.net/manual/en/class.stringable.php) interface defines `public __toString(): string` method with two underscores in name, not `toString`.

That is the main change in that pull request. Without it, there is not possible to write into a topic null value. Is not??

Steveb-p:
@Hubbitus I kinda assumed it was, my mistake.

Hubbitus:
@Steveb-p, no problem.
Can we then continue review?

Hubbitus:
Guys, could we please continue?
That is a quite trivial change I think. But very important.

Steveb-p:
> Guys, could we please continue? That is a quite trivial change I think. But very important.

Actually, it's not trivial. It breaks the contract for the interface, so semver would dictate that it should have a major version bump. Unfortunately the enqueue library, similarly to Symfony, has a shared version across all packages, so it's not exactly acceptable to just bump this one package and not the rest.

If it were to be release as a minor version, it would cause people that update to have broken builds / applications, so there is probably no solution that covers both your need (of a new functionality, kinda) and dependent applications that do have their own serializers (I'm pretty sure it would, for example, break the Symfony-Enqueue bridge for messenger).

Do you need this for your own application, or is it for your own library that is re-used by others? If it's an application or internal library, then I suggest you [overwrite the dependency with your own fork](https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository) or [rely on your own fork with `replace` entry, so it "overwrites" the enqueue](https://getcomposer.org/doc/04-schema.md#replace).

I hope that a new major version of enqueue is released eventually, but given the current state of the package (makasim, the owner, is currently occupied with projects in other languages and is not actively maintaining the lib) I cannot with good conscience promise that. It's something that probably will need to be discussed outside of this PR.

Hubbitus:
Hello.
Not introducing API/ABI breakage is a very important point, I agree. Probably indeed that can't be made in a minor version update. But what is wrong with creating new major version, list that as incompatible change (probably single), and encouraging others to step to that new version?

Symfony eventually also should update their dependencies. And said after some time (3-12 months) just drop support of current major version.

Thanks for the links about forks, of course, I already was thinking about that, but it looks like just some temporary workaround rather than the solution (in any update I will be forced to do puls, re-bases, and my releases...).

Steveb-p:
@Hubbitus the issue is that enqueue has a shared versioning, so introducing a single break should result in bumps in all packages. Which don't have any breaking changes atm (due to no small part the fact that some are simply not actively maintained).

Personally, I don't mind, but I'm technically not the owner and I cannot make statements if and when a major version bump will be done.

@makasim ?

Hubbitus:
@makasim, could you please help to schedule this into some release?

makasim:
What exactly is missing from a release ? Could you point at it ?

PR #1237, opened 2022-02-03
Fix CI

jdecool:
The CI is red and this PR fix it.

There's 3 problems:
* Thruway docker image (`docker/thruway/Dockerfile`) use `voryx/thruway` which also uses `thruway/client` which is not currently compatible with PHP 8.0. So before the library is compatible, I've fixed the PHP version to 7.4 for this container.
* `docker/Dockerfile` image is executed with PHP 7.4 & 8.0 and use `PEAR` to install some extensions. As `PR` is not fully compatible with PHP 8.0, required dependencies are compiled.
* `enqueue` libraries depends on `thruway/pawl-transport` which depend on `ratchet/pawl` (`^0.3.1`). The 0.3.x version doesn't support `guzzlehttp/psr7` 2.x branch. So we needs to restrict `guzzlehttp/psr7` to `1.x` branch

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -24,7 +24,9 @@ RUN set -x && \
         php${PHP_VERSION}-curl \
         make \
         g++ \
-        unzip
+        unzip \
+    && \
+    update-alternatives --install /usr/bin/php php /usr/bin/php${PHP_VERSION} 100
 
 ## gearman
 RUN set -x && \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -22,13 +22,20 @@ RUN set -x && \
         php${PHP_VERSION}-xml \
         php${PHP_VERSION}-mysql \
         php${PHP_VERSION}-curl \
-        libgearman-dev \
-        php-pear \
         make \
-        unzip \
+        g++ \
+        unzip
+
+## gearman
+RUN set -x && \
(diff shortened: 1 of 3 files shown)


jdecool:
Strange. I've tested it on my fork before and it was green https://github.com/jdecool/enqueue-dev/pull/1/checks

jdecool:
Could you retry the failed job ?

docker/Dockerfile
```diff
@@ -37,16 +46,19 @@ RUN set -x && \
...
-        g++ \
-    && \
     mkdir -p $HOME/librdkafka && \
     cd $HOME/librdkafka && \
     git clone https://github.com/edenhill/librdkafka.git . && \
     git checkout v1.0.0 && \
-    ./configure && make && make install && \
-    pecl install rdkafka && \
+    ./configure && make && make install
+
```
Steveb-p, line 54: Using `pecl install rdkafka-5.0.2` was not working?

Steveb-p (approved):

docker/Dockerfile
jdecool: With PHP 8.x when using `pecl` it produce a PHP warning (`Invalid argument supplied for foreach() in /usr/share/pear/PEAR/...`) that break the image build.

That's why I move it to compile it manually.

makasim:
Thanks!

PR #1239, opened 2022-02-10
Add support for Symfony 6; drop < 5.1

andrewmy:
- Allow Symfony 6 in composer deps
- Add Symfony 6 to the build matrix only for PHP 8
- Bump min php-cs-fixer to ^3.4 to allow symfony/console 6, update its config
- Bump min psr/log to 1.1
- Add TestLogger from psr/log:1 because it's removed from 2 and 3
- Fix Event Dispatcher compatibility for async events which happened to use a deprecated dispatcher configuration method
- Drop Symfony older than 5.1


Steveb-p:
<3 the commit history :sweat_smile:

andrewmy:
Currently investigating the issue with async event listeners being registered twice — either an issue with the fixture app, or Sf 6 🤔

andrewmy:
@makasim take a look pls.

I advise using "Squash merge" here to avoid letting all this commit message nonsense into master ;)

andrewmy:
Screenshot 2022-02-15 at 22 02 27
Screenshot 2022-02-15 at 22 02 37

Posting the commit message history for posterity :D

makasim:
I see some juggling with event dispatcher. 
Could dropping Symfony 4 support simplify things ? 
If so, let's do it.

makasim:
@andrewmy thank you for the great work!

andrewmy:
The new dispatcher thing was introduced in 5.1, so dropping means a bit wider scope. I'm not against it, your call :)

makasim:
Let's keep it simple. Could you set "symfony/xxx": "^5.1|^6"

Steveb-p:
If it means a new major release, then we could include #1230 in it.

andrewmy:
@Steveb-p bumping minimum deps can happen in a patch version as long as the external API is unchanged.
In this case there is a slight difference in configuration — tricky case.

andrewmy:
Used this opportunity to burn down whatever I found regarding the < 5 BC layer.

There's still the thing about one different line in fixture apps — session. Looks ugly but not warranting instant drop of anything < 5.4 I guess.

andrewmy:
Yes please, with a pretty tag on top 🙏

andrewmy:
Could we have a release? :)

PR #1250, opened 2022-05-17
Disable sleep while queue items available

mordilion:
This change is disabling the Subscription-Polling-Sleep as long there are items in queues. If there are no items in queues, then the gathering of items is not constantly penetrating the Database.

We are using it already in our production environment and it works perfectly fine without any Database issues.


jprangenbergde:
I like!

makasim:
I've hesitated with merging this because I am not sure I can spot all the possible scenario and bugs it might introduce. 

On the other hand, I dont want to be a blocker for improvements. 

I can merge if you are sure it works well. 

Please tell.

mordilion:
@makasim we are using it in a productive environment already since the 10th January 2022 and we have daily between 200.000 and 300.000 items in ~40 queues. 

Based on that I am 100% sure that it works! :)

makasim:
@mordilion thanks!

makasim:
https://github.com/php-enqueue/enqueue-dev/releases/tag/0.10.17

mordilion:
@makasim thanks for merging it :)

PR #1258, opened 2022-06-14
Fix #1257

### LINKED ISSUE #1257: [Doctrine DBAL] Invalid timestamp of published_at value
https://github.com/php-enqueue/dbal/blob/d6c6d38290415d6383ab156a9cf580a99bd05522/DbalProducer.php#L65
```
        $publishedAt = null !== $message->getPublishedAt() ?
            $message->getPublishedAt() :
            (int) (microtime(true) * 10000)
        ;
```

microtime return value in seconds so the conversion to timestamp is invalid, ex : 16551970635236 -> GMT: Monday, 5 July 2494 17:37:15.236

andrewprofile:


andrewprofile:
@makasim I described the problem in issue #1257 :)

Steveb-p:
> @makasim I described the problem in issue #1257 :)
> 
> > microtime return value in seconds so the conversion to timestamp is invalid, ex : 16551970635236 -> GMT: Monday, 5 July 2494 17:37:15.236

The issue here is that without multiplication - that you're removing - precision of stored data suffers. Messages are stored with at most 1 second precision, which makes it impossible to maintain their order & `microtime()` is equivalent to `time()`.

I think the issue you're describing is actually valid, but not because of the code. It's because we've called this property "timestamp" while we intended to store a more precise value.

andrewprofile:
@Steveb-p Sorry it's my mistake. The property published_at suggests that it will be the date, but since you operate everywhere on timestamps, I assumed that in this case also to be able to get a specific date from it. But I already understand that this is not the point, but then there is some inconsistency because other values do not operate in such precision.

Steveb-p:
> @Steveb-p Sorry it's my mistake. The property published_at suggests that it will be the date, but since you operate everywhere on timestamps, I assumed that in this case also to be able to get a specific date from it. But I already understand that this is not the point, but then there is some inconsistency because other values do not operate in such precision.

If you find any discrepancy in this or other transports feel free to open an issue or discussion (which, btw, could be enabled @makasim, so issues don't get polluted by questions). I would not outright dismiss the notion that there might be differences :sweat_smile:

makasim:
@Steveb-p Enabled discussions https://github.com/php-enqueue/enqueue-dev/discussions

andrewprofile:
@Steveb-p I meant redeliverAfter or timeToLive properties, that's why I pointed it out and assumed it was an issue :)

PR #1265, opened 2022-07-20
feat: allow rdkafka key to be set through properties

qkdreyer:
Allow serialized envelopes having `properties.key` to be able to use the rdkafka key using https://github.com/sroze/messenger-enqueue-transport


qkdreyer:
nevermind, it is already possible through through https://github.com/sroze/messenger-enqueue-transport#configure-custom-kafka-message

PR #1281, opened 2022-12-23
Add setting subscription attributes to Sns and SnsQs

andrewmy:
Not sure yet whether to add the support for attributes to `subscribe` as well, out of scope I guess.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -130,6 +130,9 @@
       "ext-mongo": "1.6.14",
       "ext-sockets": "1"
     },
-    "prefer-stable": true
+    "prefer-stable": true,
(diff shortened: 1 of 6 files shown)


andrewmy:
Hello @makasim, we're experimenting with SNS/SQS filtering, and this change is quite necessary to make it possible. Thanks!

Also please take another look at the PR from my coworker :) https://github.com/php-enqueue/enqueue-dev/pull/1283

makasim:
splitted https://github.com/php-enqueue/snsqs/commit/ebd34b8666a7107c6f2c70d792f050c7472dec95

PR #1284, opened 2023-02-08
Fix AWS SDK token parameter

andrewmy:
It looks like in recent versions the AWS SDK started enforcing parameter types, breaking our CI.

> `InvalidArgumentException: Invalid configuration value provided for "token". Expected Aws\Token\TokenInterface|Aws\CacheInterface|array|bool|callable, but got string(0) ""`

When passing `false`, it still breaks:
> `InvalidArgumentException: Token must be an instance of Aws\Token\TokenInterface, an associative array that contains "token" and an optional "expires" key-value pairs, a token provider function, or false.`

So let's just skip it in tests.


andrewmy:
@makasim could you please review? Thanks!

makasim:
@andrewmy thanks

PR #1292, opened 2023-03-23
pkg PHP 8.1 and 8.2 support

snapshotpl:
Also fix with wrong phpunit cli call with wrong parameter.


makasim:
Thanks!

snapshotpl:
@makasim can you release this one, to see how subrepositories now pass tests?

Issue #1293, opened 2023-03-29
Doctrine DBAL 3.0 and json_array


NovakHonza:
After upgrade Doctrine Dbal to 3.0 I have ended with this message:

`Unknown column type "json_array" requested. Any Doctrine type that you use has to be registered with \Doctrine\DBAL\Types\Type::addType()`

The problem is with the definition in Doctrine/mapping/Job.orm.xml in job-queue. I tried replacing the json_array with json and running the tests, but I ended up with errors in precommit hooks that I couldnt fix.

PR #1294, opened 2023-03-29
Replaced `json_array` with `json` due to Doctrine Dbal 3.0

NovakHonza:
This should be a fix for #1293


makasim:
Is it backward compatible ?

NovakHonza:
As I think about it, there might be a need for migration for existing installations? Because of the need to change the comment on the `data` column (DC2Type:json_array) to (DC2Type:json)?

Issue #940, opened 2019-08-22
Remove message after n retry attempts


YuriyChmil:
Hey!
Is there any possibility to remove the message from the queue after n attempts failed to try to handle it?
For example, if 5 times the processor couldn't handle it must remove from queue.

Also, It is possible doesn't stop the consumer when an exception occurred? In this case, I want to try redeliver this message end try to handle it one more time? I know about **REQUEUE** options, put it push message to the new queue, but I want to push to the same queue. Is it possible?

daemonfire300:
Hi there,

pushing this issue again, because:

I am using enqueue-redis and have a very simple Processor. How to I get enqueue to redeliver a Message for a later date, i.e., Message comes to processor, something temporarily did not work, let's process this message later.
What I tried:

  * Processeor::REQUEUE, but this puts a copy of the message into the queue and instantly tries to re-process it again.
  * Processor:REJECT, simply discards the message, never retries it
  * Throwing an exception in the processor, this also does not achieve the intended effect.

thaks in advance

emrahoruc:
Something like a attempt counter is needed. Any solution?

emrahoruc:
I found a solution in the php-enqueue/laravel-queue package. Uses the property field for attempts.
See:
https://github.com/php-enqueue/laravel-queue/blob/2c4ac43e0e4c65cc06406a45cab6d73991032622/src/Job.php#L62

Thank to all contributors.

PR #1301, opened 2023-05-20
fix: do not reset attemps header when message is requeue

eortiz-tracktik:
This change will allow to use `$message->getAttempts()` when `true === $message->isRedelivered()` to find how many times this message has been requeue/redelivered. Such information is useful to decide if the message should be `requeue` to retry again later or `rejected` to be completely removed from the queue in case of error.

Something similar was asked on https://github.com/php-enqueue/enqueue-dev/issues/940 but the suggestion on https://github.com/php-enqueue/enqueue-dev/issues/940#issuecomment-1017173629 is about adding another header with a similar purpose than `attempts`. 

On `\Enqueue\Redis\RedisConsumerHelperTrait::processResult()` we already increment `attempts` header, so if we don't reset it when the message is requeue we will have the accurate information about how many times the message has been consumed.

### DIFF
--- a/pkg/redis/Tests/RedisConnectionFactoryConfigTest.php
+++ b/pkg/redis/Tests/RedisConnectionFactoryConfigTest.php
@@ -195,7 +195,7 @@ class RedisConnectionFactoryConfigTest extends TestCase
             ],
         ];
 
-        //check normal redis connection for php redis extension
+        // check normal redis connection for php redis extension
         yield [
             'redis+phpredis://localhost:1234?foo=bar',
             [
@@ -218,7 +218,7 @@ class RedisConnectionFactoryConfigTest extends TestCase
             ],
         ];
 
-        //check normal redis connection for predis library
+        // check normal redis connection for predis library
         yield [
             'redis+predis://localhost:1234?foo=bar',
             [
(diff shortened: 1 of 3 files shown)


eortiz-tracktik:
Hello @makasim and @ASKozienko
Could you take a look at this proposition?
Thanks a lot.

pkg/redis/Tests/RedisConnectionFactoryConfigTest.php
```diff
@@ -264,11 +264,11 @@ public static function provideConfigs()
             ],
         ];
 
-        //check tls connection for predis library
+        // check tls connection for predis library
         yield [
             'rediss+phpredis://localhost:1234?foo=bar&async=1',
             [
-                'host' => 'tls://localhost',
+                'host' => 'localhost',
```
makasim, line 271: could you elaborate on this change ?

eortiz-tracktik: @makasim I was just trying to make the tests to pass :heavy_check_mark: 
The `master` branch is also failing on these test after your merged https://github.com/php-enqueue/enqueue-dev/pull/1297 and I don't see the tests run on this PR before being merged :thinking: 
image

makasim: that's strange indeed

PR #1304, opened 2023-06-12
[ENQUEUE] Add missing throws tag to interfaces

VincentLanglet:
HI @makasim 

When calling `sendCommand` it's useful to know the method can throws Exception, it's currently the case for the proposed implementations.
Thanks to PHPStorm or static analysis tool like PHPStan, it will help people try/catching the calls and avoid error in production.


VincentLanglet:
Friendly ping @makasim

PR #1308, opened 2023-07-12
base64 encode SNS Header message attribure

### LINKED ISSUE #1307: Unescaped SNS MessageAttributes when message headers are used
Hello,

We have recently stumbled upon an issue. We use AWS SNS/SQS setup to send messages to SNS topic and to consume them from SQS queues. We use [SnsQs transport](https://php-enqueue.github.io/transport/snsqs/) to do it. We also use message headers to add extra information to the messages we send through SNS/SQS. Additionally, we have recently started to look into using [SNS subscription filters](https://docs.aws.amazon.com/sns/latest/dg/sns-subscription-filter-policies.html) so that our services would be able to subscribe to events that they are interested in. To do so we are adding a custom message attribute to our messages using `SnsQsMessage::setMessageAttributes()` method.

However, once we enabled a Subscription filter policy on our SNS topic subscription we noticed that evens sent to the SNS topic are not getting through to the SQS queue, even tho the filtering was set up correctly. We also noticed that the value of `NumberOfNotificationsFilteredOut-InvalidAttributes` CloudWatch metric is non-zero, which lead us to believe that something is wrong with message attributes. While conducting a deeper investigation we found out that AWS SNS requires message attributes values to be escaped in case Subscription filter policies are being used. We also found [similar reports of the issue](https://repost.aws/questions/QUyUJ6EvX6So2I1MAS-c-dmg/an-sns-message-with-a-double-quote-in-any-attribute-is-filtered-out-is-this-a-bug) on AWS forum.

We believe the problem is that the [value of json_encode](https://github.com/php-enqueue/enqueue-dev/blob/4a78c2eab5caafe7cf1c3682592791db5aa12e65/pkg/sns/SnsProducer.php#L54) is not being escaped using `addslashes()` function. We modified the enqueue library code locally and found that adding `addslashes()` works. We also noticed that the addition of `stripslashes()` would be required in [SnsQsConsumer](https://github.com/php-enqueue/enqueue-dev/blob/4a78c2eab5caafe7cf1c3682592791db5aa12e65/pkg/snsqs/SnsQsConsumer.php#L127) class to remove the previously added slashes.

However, we believe such a change would be breaking backward compatibility as the messages that have been previously sent without `addslashes()` would become unconsumable if `stripslashes()` is introduced.

Another potential solution could be to bease64 encode the result of `json_encode()` and on message retrieval to base64 decode it only if it does not start with `[` character (it was not base64 encoded). This would at least allow previously sent messages to be consumed without issues, however, the downside of such approach is that the raw message becomes less readable.

We would like to make a contribution to fixing this issue but currently are stuck due to the BC break issue. Perhaps maintainers or other contributors to this project have ideas on how to proceed further?

Below is the code we used to replicate the issue:

Send message:
```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

use Enqueue\SnsQs\SnsQsConnectionFactory;

$factory = new SnsQsConnectionFactory([
    'key'    => 'AAAAAAAAAAAAAAAAAAAA',
    'secret' => 'kkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkk',
    'region' => 'us-east-2',
]);

$context = $factory->createContext();

$inTopic = $context->createTopic('in');
$context->setTopicArn($inTopic, 'arn:aws:sns:us-east-2:000000000000:test-topic');

$message = $context->createMessage('Hello world 123!');
$message->setHeader('My-Custom-Header', 'my "value');

$message->setMessageAttributes([
    'EventName' => [
        'DataType'    => 'String',
        'StringValue' => 'events/test/dummy',
    ]
]);

$context->createProducer()->send($inTopic, $message);
```

Get message:
```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

use Enqueue\SnsQs\SnsQsConnectionFactory;

$factory = new SnsQsConnectionFactory([
    'key'    => 'AAAAAAAAAAAAAAAAAAAA',
    'secret' => 'kkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkk',
    'region' => 'us-east-2',
]);

$context = $factory->createContext();

$out1Queue = $context->createQueue('example-queue');
$consumer = $context->createConsumer($out1Queue);

$message = $consumer->receiveNoWait();

var_dump($message);

if (null !== $message) {
    $consumer->acknowledge($message);
}
```

Our `example-queue` SQS queue is subscribed to `test-topic` SNS topic and the following Subscription filter policy is used:
```json
{
  "EventName": [
    "events/test/dummy"
  ]
}
```

martinssipenko:
Fixes #1307

### DIFF
--- a/pkg/sns/SnsProducer.php
+++ b/pkg/sns/SnsProducer.php
@@ -51,7 +51,7 @@ class SnsProducer implements Producer
             'MessageAttributes' => [
                 'Headers' => [
                     'DataType' => 'String',
-                    'StringValue' => json_encode([$message->getHeaders(), $message->getProperties()]),
+                    'StringValue' => base64_encode(json_encode([$message->getHeaders(), $message->getProperties()])),
                 ],
             ],
             'TopicArn' => $topicArn,
--- a/pkg/sns/Tests/SnsProducerTest.php
+++ b/pkg/sns/Tests/SnsProducerTest.php
@@ -138,7 +138,7 @@ class SnsProducerTest extends TestCase
             'MessageAttributes' => [
                 'Headers' => [
                     'DataType' => 'String',
-                    'StringValue' => '[{"hkey":"hvaleu"},{"key":"value"}]',
+                    'StringValue' => base64_encode('[{"hkey":"hvaleu"},{"key":"value"}]'),
                 ],
             ],
             'TopicArn' => 'theTopicArn',
@@ -189,7 +189,7 @@ class SnsProducerTest extends TestCase
             'MessageAttributes' => [
                 'Headers' => [
                     'DataType' => 'String',
-                    'StringValue' => '[[],[]]',
+                    'StringValue' => base64_encode('[[],[]]'),
                 ],
                 'Foo' => [
                     'DataType' => 'String',
--- a/pkg/snsqs/SnsQsConsumer.php
+++ b/pkg/snsqs/SnsQsConsumer.php
@@ -124,7 +124,11 @@ class SnsQsConsumer implements Consumer
                 }
 
                 if (isset($data['MessageAttributes']['Headers'])) {
-                    $headersData = json_decode($data['MessageAttributes']['Headers']['Value'], true);
+                    $rawHeaderData = $data['MessageAttributes']['Headers']['Value'];
+                    if (isset($rawHeaderData[0]) && '[' !== $rawHeaderData[0]) {
+                        $rawHeaderData = base64_decode($rawHeaderData);
+                    }
+                    $headersData = json_decode($rawHeaderData, true);
 
                     $message->setHeaders($headersData[0]);
                     $message->setProperties($headersData[1]);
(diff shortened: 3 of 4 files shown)


rolandsusans (approved):

martinssipenko:
This solution would not work in cases where there are multiple distinct services using enqueue for cross-service communication via messages because this change would introduce bease64 encoding of messages, but if there are other services using an older version of this library, that can not base64 decode messages, the message consumption would break.

pkg/sns/SnsProducer.php
```diff
@@ -51,7 +51,7 @@ public function send(Destination $destination, Message $message): void
             'MessageAttributes' => [
                 'Headers' => [
                     'DataType' => 'String',
-                    'StringValue' => json_encode([$message->getHeaders(), $message->getProperties()]),
+                    'StringValue' => base64_encode(json_encode([$message->getHeaders(), $message->getProperties()])),
```
rolandsusans, line 54: Maybe we can leave this without encoding base64 for now (keep changes not breaking). 
In the next major release, we can add encode part. 🤔

PR #1310, opened 2023-08-09
Set return type to flush function in RdkafkaProducer to be compatible to php-rdkafka

felippeduarte:
Looking at the rdkafka docs (https://arnaud.le-blanc.net/php-rdkafka-doc/phpdoc/rdkafka.flush.html) flush returns an integer instead of void in Producer class. This PR sets the same return type.


williamtcastro (approved):

felippeduarte:
hey @makasim ,  could you please take a look?

PR #1312, opened 2023-08-17
Add `void` return type in DoctrineSchemaCompilerPass and MessageQueueCollector

ruudk:
Fixes these deprecations on Symfony 6.3:

```
Method "Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface::process()" might add "void" as a native return type declaration in the future. Do the same in implementation "Enqueue\Doctrine\DoctrineSchemaCompilerPass" now to avoid errors or add an explicit @return annotation to suppress this message.
Method "Symfony\Component\HttpKernel\DataCollector\DataCollectorInterface::collect()" might add "void" as a native return type declaration in the future. Do the same in implementation "Enqueue\Bundle\Profiler\MessageQueueCollector" now to avoid errors or add an explicit @return annotation to suppress this message.
```

### DIFF
--- a/pkg/enqueue-bundle/Profiler/MessageQueueCollector.php
+++ b/pkg/enqueue-bundle/Profiler/MessageQueueCollector.php
@@ -8,6 +8,9 @@ use Throwable;
 
 class MessageQueueCollector extends AbstractMessageQueueCollector
 {
+    /**
+     * @return void
+     */
     public function collect(Request $request, Response $response, Throwable $exception = null)
     {
         $this->collectInternal($request, $response);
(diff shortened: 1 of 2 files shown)


ruudk:
@makasim This one is ready for review.

makasim:
Unfortunately, I cannot merge the red build, If it makes to the main it will stay there for a long time, making the whole idea of testing and CI useless.

ruudk:
Main is already red: https://github.com/php-enqueue/enqueue-dev/commits/master

The error on this PR is the same, so that's what I mean.

Obviously this PR doesn't create new errors 😅

ruudk:
I give up. The errors in this PR are unrelated to my changes.

ruudk:
This fails because this project uses PHPStan 0.12, which is outdated. It should be updated to PHPStan ^1.8.

PR #1355, opened 2024-08-13
GPS: revert the attributes and use the headers instead.

p-pichet:
I found that i was go in wrong direction. 
The GPS attributes are the Interop Message headers. 

I revert to use the `headers` and remove the `attributes`

### DIFF
--- a/pkg/gps/GpsMessage.php
+++ b/pkg/gps/GpsMessage.php
@@ -25,11 +25,6 @@ class GpsMessage implements Message, \JsonSerializable
     private $headers;
 
     /**
-     * @var array
-     */
-    private $attributes;
-
-    /**
      * @var bool
      */
     private $redelivered;
@@ -39,12 +34,11 @@ class GpsMessage implements Message, \JsonSerializable
      */
     private $nativeMessage;
 
-    public function __construct(string $body = '', array $properties = [], array $headers = [], array $attributes = [])
+    public function __construct(string $body = '', array $properties = [], array $headers = [])
     {
         $this->body = $body;
         $this->properties = $properties;
         $this->headers = $headers;
-        $this->attributes = $attributes;
 
         $this->redelivered = false;
     }
@@ -157,7 +151,6 @@ class GpsMessage implements Message, \JsonSerializable
             'body' => $this->getBody(),
             'properties' => $this->getProperties(),
             'headers' => $this->getHeaders(),
-            'attributes' => $this->getAttributes(),
         ];
     }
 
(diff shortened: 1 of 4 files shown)


Steveb-p:
@p-pichet the code you added, with explicit attributes, has just been released.

Technically, now you're breaking the changes you just introduced. Backwards compatibility breaks are a big no-no for libraries.

You have to explain it precisely why you want to go back, and what caused you to change your mind - links to resources, discussion that proves the issue, code samples, and so on.

p-pichet:
@Steveb-p, when I use it on my project, I don't have access to attributes in the consume message. 

I check on the [InteropQueue](https://github.com/queue-interop/queue-interop/blob/master/src/Message.php) and found that I make a mistake of global compatibility. 

Base on this comment : 

> The header contains fields used for message routing and identification;

The GPS attributes are simply headers. 

By using a custom property `attributes`, we break the possibility to easily modify the transport by modify configuration.

What can I propose to avoid breaking change is this : 
On the class `GPSMessage` : 
- keep getter and setter of attributes and mark them deprecated and remove them in the V1 
- on the getHeaders, merge the attributes with headers. If the both array have the same key, we keep the one in headers
On the GPSProducer, we keep my refacto to send headers

What do you think ?

pkg/gps/GpsMessage.php
```diff
@@ -181,11 +180,17 @@ public function setNativeMessage(?GoogleMessage $message = null): void
         $this->nativeMessage = $message;
     }
 
+    /**
+     * @deprecated This will be removed in the V1, use headers instead
```
makasim, line 184: nah, I think we can break it right away. I don think there are many people that started using it.

p-pichet: thanks just drop the last commit.

p-pichet:
@makasim I revert it. 

@Steveb-p, i'm sorry. That was when I use it with sroze/messenger-enqueue-tranposrt that I saw I don't respect the 'standard' of interop enqueue.

Steveb-p:
@p-pichet Don't worry :) It's just that when you're providing software for someone else, suddenly changes you can make in your own project that you were able to do are forbidden :)

If I can offer you a bit of advice - when working on a project and using open source libraries that you need to change to work, you can create your own forks and install them in place of the original.

See https://getcomposer.org/doc/05-repositories.md#repository .

For example, you can fork both enqueue-dev and sroze adapter, and do the changes that you need to. They can both be installed using VCS...
```json
{
    "repositories": [
        {
            "type": "vcs",
            "url": "git@github.com:my-fork/enqueue-dev.git"
        }
    ]
}
```
...or even local paths:
```
{
    "repositories": [
        {
            "type": "path",
            "url": "../enqueue-dev-fork"
        }
    ]
}
```

And you can install branches pretending to be stable releases:
```
composer require "php-enqueue/enqueue-dev as 0.11"
```
(at least I think that's the syntax)

This is also one way of getting the features you need in your project, while waiting for your PRs to be approved.

p-pichet:
@Steveb-p thanks for the tips. 

@makasim thanks for the merge. When do you think you will create the new release ?

makasim:
https://github.com/php-enqueue/enqueue-dev/releases/tag/0.10.22

PR #1369, opened 2025-01-09
use explicitly nullable parameter types (implicit deprecated in php 8.4)

slepic:
> PHP supports nullable type declarations as of PHP 7.1 with the [?T syntax](https://wiki.php.net/rfc/nullable_types), and with the [T|null syntax](https://wiki.php.net/rfc/union_types_v2) as of PHP 8.0 when generalized support for union types was added.

https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

enqueue packages declares to require at least php ^7.4, therefore the ?T syntax is used here.

I believe that many of those parameters dont actually need a default value, but I leave that out because it might be BC if someone is actually calling those setters with no params.

### DIFF
--- a/pkg/amqp-bunny/AmqpConsumer.php
+++ b/pkg/amqp-bunny/AmqpConsumer.php
@@ -48,7 +48,7 @@ class AmqpConsumer implements InteropAmqpConsumer
         $this->flags = self::FLAG_NOPARAM;
     }
 
-    public function setConsumerTag(string $consumerTag = null): void
+    public function setConsumerTag(?string $consumerTag = null): void
     {
         $this->consumerTag = $consumerTag;
     }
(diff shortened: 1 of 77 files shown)


Steveb-p:
I think we cannot do that. Changing argument types by widening them for classes that can potentially be extended - as we do not prohibit it - would be a breaking change (syntax error) if any of those methods is redeclared.

This deprecation can only be removed in major version, and as you've noted majority of those do not actually allow `null` values.

slepic:
@Steveb-p it's actually not widening the types, it just uses more uptodate syntax for the same

see here if you extend a class which has `?T $param = null`, and change it to `T $param = null`, it is totally ok

https://phpstan.org/r/0f1a0e2c-47cd-4a0a-9314-35f1e5dfe1d7

PR #1371, opened 2025-01-11
Fixing risky tests

### LINKED ISSUE #1370: Fix risky unit tests
When running the tests for this project, a lot of noise is generated between deprecation warnings and risk tests that don't have any assertions.

This makes it difficult to read the log, it would be nice if this were clean up to fix all the risky tests.

JimTools:
Making tests that have no assertions with the annotation to suppress phpunits warning around risky tests

closes #1370

### DIFF
--- a/pkg/amqp-bunny/Tests/AmqpConsumerTest.php
+++ b/pkg/amqp-bunny/Tests/AmqpConsumerTest.php
@@ -27,14 +27,6 @@ class AmqpConsumerTest extends TestCase
         $this->assertClassImplements(Consumer::class, AmqpConsumer::class);
     }
 
-    /**
-     * @doesNotPerformAssertions
-     */
-    public function testCouldBeConstructedWithContextAndQueueAsArguments()
-    {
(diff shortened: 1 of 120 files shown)


JimTools:
To be honest I don't think a lot of these tests bring value, for the most part, they just test the constructor works can get a new instance.

makasim:
At the time of writing (6-8 years ago), it seemed like a good idea. Now, I agree with you that such tests bring no value. 

> To be honest I don't think a lot of these tests bring value, for the most part, they just test the constructor works can get a new instance.

JimTools:
I’ll work on updating this to remove tests  that have little to no value.

As for the CS and static analysis I’ll raise another PR to fix this.

JimTools:
Fixed CS issues in #1372

JimTools:
Fixes for static analysis are in #1373, it's best if this gets merged after the cs fixes. 

@makasim how would you like to handle the 3 PR's? I plan to create more fixing deprecations and warnings in the tests. and possibly upgrade PHPUnit to 10.x

PR #1382, opened 2025-04-18
Fixing CI

JimTools:
There's an issue with the downstream PPA "ondrej/php" where is cannot find the gearman package.

Refactoring Dockerfile to use standard base images as a workaround

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,71 +1,34 @@
 ARG PHP_VERSION=8.2
-FROM makasim/nginx-php-fpm:${PHP_VERSION}-all-exts
+FROM php:${PHP_VERSION}-alpine
 
 ARG PHP_VERSION
 
-## libs
-RUN set -x && \
-    apt-get update && \
-    apt-get install -y --no-install-recommends --no-install-suggests \
-        wget \
-        curl \
-        openssl \
-        ca-certificates \
-        nano \
-        netcat \
-        php${PHP_VERSION}-dev \
-        php${PHP_VERSION}-redis \
-        php${PHP_VERSION}-pgsql \
-        git \
-        python \
-        php${PHP_VERSION}-amqp \
-        php${PHP_VERSION}-xml \
-        php${PHP_VERSION}-mysql \
-        php${PHP_VERSION}-curl \
-        php${PHP_VERSION}-mongodb \
-        php${PHP_VERSION}-mbstring \
-        make \
-        g++ \
-        unzip \
-    && \
-    update-alternatives --install /usr/bin/php php /usr/bin/php${PHP_VERSION} 100
+RUN --mount=type=cache,target=/var/cache/apk apk add --no-cache $PHPIZE_DEPS \
+    libpq-dev \
+    librdkafka-dev \
+    rabbitmq-c-dev \
+    linux-headers && \
+    apk add --no-cache --repository=https://dl-cdn.alpinelinux.org/alpine/edge/testing \
+    gearman-dev
 
-## gearman
-RUN set -x && \
-    apt-get install -y --no-install-recommends --no-install-suggests \
-        libgearman-dev \
-    && \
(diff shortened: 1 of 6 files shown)


JimTools:
I've managed to get the majority of the tests to pass. The one which I can't work out is:

`TypeError: is_dir(): Argument #1 ($filename) must be of type string, int given`

This seems deep in Syfoney when it caches the config in `/tmp` and makes a proxy call.

### Notable changes

With the Dockerfile changes, it may be beneficial to cache the build steps or push the image to a repository for faster build time and less waiting on tests

JimTools:
Hi @makasim, I've managed to get the test to pass 🎉 

However, I did have to drop the lower supported versions of Syfoney to get around  `TypeError: is_dir(): Argument #1 ($filename) must be of type string, int given`

makasim:
Thank you very much, keep CI green is very important job.

> I used to support Bob Martin’s early ideas, but over time, it’s become increasingly difficult. On this point, I couldn’t disagree more.

That's okay.

docker/Dockerfile
```diff
@@ -1,76 +1,39 @@
...
-    mkdir -p $HOME/php-rdkafka && \
-    cd $HOME/php-rdkafka && \
-    git clone https://github.com/arnaud-lb/php-rdkafka.git . && \
-    git checkout 5.0.1 && \
-    phpize && ./configure && make all && make install && \
-    echo "extension=rdkafka.so" > /etc/php/${PHP_VERSION}/cli/conf.d/10-rdkafka.ini && \
-    echo "extension=rdkafka.so" > /etc/php/${PHP_VERSION}/fpm/conf.d/10-rdkafka.ini
-
-COPY ./php/cli.ini /etc/php/${PHP_VERSION}/cli/conf.d/1-dev_cli.ini
+RUN --mount=type=cache,target=/var/cache/apk apk add --no-cache $PHPIZE_DEPS \
```
makasim, line 6: will this work on macos ?

JimTools: Yep this is a feature of docker it’s not platform depend.

https://docs.docker.com/build/cache/optimize/#use-cache-mounts

JimTools:
> I used to support Bob Martin’s early ideas, but over time, it’s become increasingly difficult. On this point, I couldn’t disagree more.

I still think there is wisdom in what he has said, but like most things, take it with a pinch of salt

docker/Dockerfile
makasim: TIL, thanks!

makasim:
LGTM, happy to merge once green

JimTools:
That test seems flaky, I'll take a look at it tonight.

```
/mqdev # ./bin/phpunit --filter=RedisSubscriptionConsumerConsumeFromAllSubscribedQueuesTest --repeat 100
PHPUnit 9.6.22 by Sebastian Bergmann and contributors.

.......................................F.......................  63 / 100 ( 63%)
.....................................                           100 / 100 (100%)

Time: 01:59.564, Memory: 38.00 MB

There was 1 failure:

1) Enqueue\Redis\Tests\Spec\RedisSubscriptionConsumerConsumeFromAllSubscribedQueuesTest::test
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => 'fooBody'
-    1 => 'barBody'
 )

/mqdev/vendor/queue-interop/queue-spec/src/SubscriptionConsumerConsumeFromAllSubscribedQueuesSpec.php:63

FAILURES!
Tests: 100, Assertions: 299, Failures: 1.

JimTools:
Looks like that test case extends `queue-interop/queue-spec`. I've re-run the test and all green this time.

I'm not sure what the best course of action is.

https://github.com/queue-interop/queue-spec/blob/master/src/SubscriptionConsumerConsumeFromAllSubscribedQueuesSpec.php

PR #1390, opened 2025-05-20
feat: Make serializer configurable via YAML configuration

lenchester:
This change enables users to specify custom serializers with configuration  
options in their Enqueue bundle YAML configuration, improving flexibility  
when working with different message formats.

- Add support for configuring serializers through YAML in RdKafkaContext  
- Allow serializer specification as a class name, array with options, or instance

### DIFF
--- a/pkg/rdkafka/RdKafkaContext.php
+++ b/pkg/rdkafka/RdKafkaContext.php
@@ -16,7 +16,6 @@ use Interop\Queue\Producer;
 use Interop\Queue\Queue;
 use Interop\Queue\SubscriptionConsumer;
 use Interop\Queue\Topic;
-use InvalidArgumentException;
 use RdKafka\Conf;
 use RdKafka\KafkaConsumer;
 use RdKafka\Producer as VendorProducer;
@@ -56,34 +55,6 @@ class RdKafkaContext implements Context
         $this->kafkaConsumers = [];
         $this->rdKafkaConsumers = [];
         $this->configureSerializer($config);
-    }
-
-    /**
-     * @param array $config
-     * @return void
-     */
-    private function configureSerializer(array $config): void
-    {
-        if (!isset($config['serializer'])) {
-            $this->setSerializer(new JsonSerializer());
-            return;
-        }
-
-        if (is_string($config['serializer'])) {
-            $this->setSerializer(new $config['serializer']());
-        } elseif (is_array($config['serializer']) && isset($config['serializer']['class'])) {
-            $serializerClass = $config['serializer']['class'];
-            $serializerOptions = $config['serializer']['options'] ?? [];
-            if (!empty($serializerOptions)) {
-                $this->setSerializer(new $serializerClass($serializerOptions));
-            } else {
-                $this->setSerializer(new $serializerClass());
-            }
-        } elseif ($config['serializer'] instanceof Serializer) {
-            $this->setSerializer($config['serializer']);
-        } else {
-            throw new InvalidArgumentException('Invalid serializer configuration');
-        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


pkg/rdkafka/RdKafkaContext.php
```diff
@@ -54,8 +55,35 @@ public function __construct(array $config)
...
+            if (!empty($serializerOptions)) {
+                $this->setSerializer(new $serializerClass($serializerOptions));
+            } else {
+                $this->setSerializer(new $serializerClass());
+            }
+        } elseif ($config['serializer'] instanceof Serializer) {
+            $this->setSerializer($config['serializer']);
+        } else {
+            throw new InvalidArgumentException('Invalid serializer configuration');
+        }
```
Steveb-p: Personally, I'd prefer something like this, with more verbosity provided for invalid configuration values.

`$serializerClass` "normalization" could be extracted to a separate private method. Similarly, exception message/instance creation.

Also, added check that the instance we are creating is actually of the right class. Usually object creation is a domain of an app, but in rare circumstance that somehow user could control what ends up in `$config['serializer']`, a malicious user could for example create PHP subprocesses that would allow backdoor access to a server. Unlikely, but it doesn't hurt to check.

```suggestion
        if ($config['serializer'] instanceof Serializer) {
            $this->setSerializer($config['serializer']);
            return;
        }

        if (is_string($config['serializer']) {
            $serializerClass = $config['serializer'];
        } elseif (is_array($config['serializer'] && isset($config['serializer']['class']) {
            $serializerClass = $config['serializer']['class'];
        } else {
            throw new InvalidArgumentException(sprintf(
                'Invalid serializer configuration. Expected "%s" option to be string, array with "class" property, or %s instance. Received %s instead.',
                'serializer',
                Serializer::class,
                get_debug_type($config['serializer']),
            );
        }

        if (!class_exists($serializerClass) || !is_a($serializerClass, Serializer, true)) {
            throw new InvalidArgumentException(sprintf(
                'Invalid serializer configuration. Expected "%s" option to be string, array with "class" property, or %s instance. Received %s instead.',
                'serializer',
                Serializer::class,
                $serializerClass,
            );
        }

        $serializerOptions = $config['serializer']['options'] ?? [];
        $this->setSerializer(new $serializerClass($serializerOptions));
```

lenchester: @Steveb-p  I have refactored the code according to your suggestions and have added default fallback in order not to break backwards compatibility.

pkg/rdkafka/RdKafkaContext.php
```diff
@@ -180,6 +179,58 @@ public static function getLibrdKafkaVersion(): string
         return "$major.$minor.$patch";
     }
 
+    /**
+     * @return void
+     *              JsonSerializer should be the default fallback if no serializer is specified
+     */
```
Steveb-p: No need to re-specify the return type, since it's already present in the PHP declaration.

As for the description, I'm not sure it's worthwile - but nothing against it either.

```suggestion
    /**
     * JsonSerializer is the default fallback, if no serializer is specified.
     */
```

pkg/rdkafka/Tests/RdKafkaContextTest.php
```diff
@@ -36,6 +36,34 @@ public function testShouldSetJsonSerializerInConstructor()
         $this->assertInstanceOf(JsonSerializer::class, $context->getSerializer());
     }
 
+    public function testShouldUseStringSerializerClassFromConfig()
```
Steveb-p, line 39: Consider adding unit test that checks `assertSame` instance when serializer object is passed. If you are able you might want to check constructor options with method call assertions.

Steveb-p (approved):

pkg/rdkafka/Tests/RdKafkaContextTest.php
lenchester: I'm not sure what you mean by checking constructor options with method call assertions. Could you please explain in more detail? Do you mean  I should check how `configureSerializer` is being called from constructor of RdKafkaContext? I think, I think, I cannot assert private methods in Mocks. 

P.S sorry for delayed response. I took a break from coding. Now I am really up to finish this.

lenchester:
@Steveb-p can you review the PR?

pkg/rdkafka/Tests/RdKafkaContextTest.php
Steveb-p: The test that you added in [ff9de7f](https://github.com/php-enqueue/enqueue-dev/pull/1390/commits/ff9de7f4241fb268f144fbf6814220d476f575b9) is the exact thing I was looking for :)

Steveb-p (approved):

lenchester:
Hmm, I’m not sure whether the error was caused by my changes.

Steveb-p:
> Hmm, I’m not sure whether the error was caused by my changes.

Seems like it was a one-off failure in one job in tests.

GhassenRjab:
Hey,
Thank you for working on this!

I am working in a Symfony app where Kafka is configured only with yaml files (enqueue and messenger).

I hope this feature gets added soon !

In the meantime, how can I configure a custom serializer without yaml? Can you help me please?

ovovov219-hash:
Hi team, thank you for implementing this feature! Would it be possible to cut a new release with it? We’re currently blocked on this. Appreciate your work on this.

ovovov219-hash:
@makasim Is it possible to release a new version? Thank you.

lenchester:
This PR is a bit stale. Is there anything I could help to get it merged and released? @makasim

PR #1391, opened 2025-06-21
add symfony 7 support + adjust CI for simple-client package

LordSimal:
Refs: https://github.com/php-enqueue/simple-client/pull/2


pkg/simple-client/composer.json
```diff
@@ -10,7 +10,7 @@
         "enqueue/enqueue": "^0.10",
         "queue-interop/amqp-interop": "^0.8.2",
         "queue-interop/queue-interop": "^0.8",
-        "symfony/config": "^5.4|^6.0"
+        "symfony/config": "^5.4|^6.0|^7.0"
```
makasim: Could you please drop 5.4 version? I think keeping current and previous is enough.

PR #1394, opened 2025-07-16
Improve log extension to avoid unique messages

SpartakusMd:
Currently the log extension is adding the processed message body to the log message. This creates unique logs, and duplicates the info as it's added in the log context. It's not an issue per se, but if you use Sentry (or similar tools), you receive issues for each message handled with error.

What do you think about excluding it?

### DIFF
--- a/pkg/enqueue/Consumption/Extension/LogExtension.php
+++ b/pkg/enqueue/Consumption/Extension/LogExtension.php
@@ -30,7 +30,7 @@ class LogExtension implements StartExtensionInterface, MessageReceivedExtensionI
     {
         $message = $context->getMessage();
 
-        $context->getLogger()->debug("Received from {queueName}", [
+        $context->getLogger()->debug('Received from {queueName}', [
             'queueName' => $context->getConsumer()->getQueue()->getQueueName(),
             'redelivered' => $message->isRedelivered(),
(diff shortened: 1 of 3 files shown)


Steveb-p:
Sending `debug` level logs to Sentry is, in general, a bad idea. Even with a on-premise setup it will simply generate a ton of useless logs (unless you make more complex decision making about what is sent and what isn't based on maybe channel settings).

That said, I agree that if it's in context, it could be removed from the main message string.

SpartakusMd:
The thing is that it's an `error` level log if the processing of the message was not successful. And you cannot easily filter messages that are sent to Sentry. Please see https://github.com/php-enqueue/enqueue-dev/pull/1394/files#diff-61f6a09a5a7e5b04b1179fa9524988b42ac12eace842080d2397c699af81a373L63

Steveb-p:
Right, I see now :) Yeah, definitely looks good to me, just need tests adjusted to the new variant.

SpartakusMd:
Sure, thing. Should be fine now.

SpartakusMd:
Does it look good? Is there anything else I could help to get it merged and released?

SpartakusMd:
@Steveb-p any way to help this move forward?

SpartakusMd:
@makasim could you review this when you have some time?

